PR #1, opened 2017-06-05
Initial version of theme.

jonparrott:
Extracted from CPython @ 1bced5656

Added a few more theme config values, split the donate footer into its own template, and removed CPython-docs sepcific JavaScript and HTML.


jonparrott:
@brettcannon FYI I have a working version of the CPython docs that use this theme (and removes duplicated code), should I send that over to CPython after this is merged? :)

brettcannon:
@jonparrott sure

README.rst
```diff
@@ -0,0 +1,4 @@
+Python Docs Sphinx Theme
+=========================
+
+This is the theme for the Python documentation
```
ncoghlan: The note from the python-dev discussion can just be added here as a second paragraph:

> Note that when adopting this theme, you're also borrowing an element of the trust and credibility established by the CPython core developers over the years. That's fine, and you're welcome to do so for other Python community projects if you so choose, but please keep in mind that in doing so you're also choosing to accept some of the responsibility for maintaining that collective trust.

setup.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
+
+import io
+
+from setuptools import setup
+
+# README into long description
+with io.open('README.rst', encoding='utf-8') as readme_file:
```
ncoghlan: Are we planning to support installing this theme for use in Py2 docs builds? If not, we can just use the open() builtin directly here, rather than going through the `io` module.

README.rst
jonparrott: Done, thank you. :)

setup.py
jonparrott: Although this is technically for the Python 3 docs, using `io.open` over `open` is pretty cheap for us, especially considering this page contains almost no Python code. :)

jonparrott:
@ncoghlan please take another look when you get a chance. :)

jonparrott:
Thanks!

PR #2, opened 2017-06-09
Add installation instructions

ncoghlan:
While the name *has* been reserved on PyPI, I'm not sure it actually
makes sense to publish the theme that way, so these instructions
just recommend installing it directly from `git`.


jonparrott:
LGTM.

I reserved the name, but agree that might not be the best way to distribute this at the moment. :)

PR #25, opened 2019-02-13
Change the font-family for the parameters of a function

matrixise:
Change the font-family for the parameters of a function.

It's related to this issue: https://bugs.python.org/issue35986


theacodes:
Thanks!

Issue #23, opened 2018-08-21
The help text for [>>>] toggle always in English


jfine2358:
See, for example, https://docs.python.org/fr/3/tutorial/introduction.html. It applies to all interactive code examples.

This is https://bugs.python.org/issue34452, transferred here, as a third party issue. It is related to 
[PEP 545 - Python Documentation Translations](https://www.python.org/dev/peps/pep-0545/).

Here's what I've found. The purpose of python_docs_theme/static/copybutton.js is
> Add a [>>>] button on the top-right corner of code samples to hide the >>> and ... prompts and the output and thus make the code copyable.

To do this it contains code
```javascript
    var hide_text = 'Hide the prompts and output';
    var show_text = 'Show the prompts and output';
```
Somehow, this text needs to be internationalised.

theacodes:
Tricky. Not sure how to internationalize text within a JS file.

We could see about moving that text into the Jinja templates and use JS to show/hide it.

jfine2358:
Thank you for your prompt attention. It's not something I've done, either. I looked at `view-source:https://docs.python.org/3/tutorial/introduction.html`. It contains
```html
<script type="text/javascript" src="../_static/copybutton.js"></script>
<script type="text/javascript" src="../_static/switchers.js"></script>
```
Perhaps best not to change this. Instead load an extra file, something like `static/fr_text.js`. that contains all the language variable text. This file should I think be part of the FRench translation of the docs (or generated from such). Once you've got that, use it in `copybutton.js`.
There are libraries and tools associated with GNU gettext, that will help you with this. I hope this helps.

theacodes:
I've honestly never done translation stuff, but I'm happy to give it a shot once I find some time.

In the meantime, it would be awesome if anyone with specific experience here can help.

jfine2358:
This might be helpful.
https://docs.djangoproject.com/en/2.1/topics/i18n/translation/#internationalization-in-javascript-code

terryjreedy:
I would talk to the people doing doc translations: https://devguide.python.org/experts/#documentation-translations.  Discussion on https://mail.python.org/mailman/listinfo/doc-sig

ccwang002:
While is it hard to add translation directly into the javascript file, I was wondering if it is possible to add the strings in a HTML file, and use jQuery to fetch the translated string back. 

Sphinx has the ability to extract sentences from HTML. For example, we could add the strings into `layout.html`, on which all the HTML pages are based:

```django
[...]

PR #27, opened 2019-02-17
Make copybutton messages translatable

ccwang002:
This PR implements #23 which make the messages of copybutton (`>>>`) translatable.

When I tested my change using the master branch of Python (branch 3.7 doesn't seem to use python-docs-theme), I couldn't get it read my modified `layout.html`. However, if I manually introduced the change, the pot files would include the new strings and the translation worked.

I think it is better to have someone more familiar with the project structure of PEP 545 to test my implementation before it is merged.

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -57,4 +57,11 @@
 
     {% trans sphinx_version=sphinx_version|e %}Created using <a href="http://sphinx.pocoo.org/">Sphinx</a> {{ sphinx_version }}.{% endtrans %}
     </div>
+    {# Translated messages used by copybutton #}
+    {% if builder != "htmlhelp" and not embedded %}
+    <div class="copybutton-msg" style="display: none">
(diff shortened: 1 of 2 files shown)


JulienPalard:
There's a little issue, the command gathering translations strings:
```
sphinx-build -Q -b gettext -D gettext_compact=0 Doc pot/
```

won't pick it up, so translation teams won't see it as a sentence to translate. I don't have the time *right now* to check if there's a simple way to solve this, maybe Sphinx has a way to look for translations strings in configured theme?

theacodes:
These resources might be helpful:

1. https://github.com/rtfd/sphinx_rtd_theme/issues/403
2. https://github.com/rtfd/sphinx_rtd_theme/pull/405

septatrix:
I think I looked into this in the past and it was possible to directly translate the strings in the javascript file

hugovk:
@tomasr8 Can this be closed now https://github.com/python/python-docs-theme/pull/225 is merged?

tomasr8:
I believe so, the button title is now properly i18n'd.

PR #28, opened 2019-05-08
[WIP] Add warning about old versions

JulienPalard:
This is a draft on how we could provide a link to newer versions:

deprecated

Co-Authored-By: @matrixise 

It lacks a few little things:

- I have not tested it yet, my laptop here is too slow to build multiple languages
- I think that over the pure html link (good for SEO) we could add a layer of javascript to test the existence of the target page and fallbacking to `/` instead of linking to a 404.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,3 +13,31 @@ trust.
 To use the theme, install it into your docs build environment via ``pip``::
 
     pip install python-docs-theme
+
+
+Features
+========
+
+Outdated redirection
+--------------------
+
+You can define the ``outdated_message`` and ``outdated_link_text``
+variables to show a red banner on each page redirecting to the "latest"
+version.
+
+The ``outdated_message`` comes first, the ``outdated_link_text`` comes
+afterwards and is the one user can click. When clicking on
+``outdated_link_text``, they will be redirected to the same page
+stripped from any prefix.
+
+Meaning if they're on ``/2.7/``, they'll go to ``/``, if they're on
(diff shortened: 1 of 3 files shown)


theacodes:
@JulienPalard this looks great to me so far. Are you planning on handling the two remaining items (testing and clever javascript) or are you wanting someone on this end to take it on?

matrixise:
Hi @theacodes 

If you want to work on the JavaScript, you could see the code of the switchers.

Thank you

I’ll review this PR

theacodes:
Honestly looking at the way `swichers.js` is implemented in cpython, it probably make sense to just move all of this logic there. If anything, put a new block here maybe (`block pre_body_area`) to allow the cpython configuration to inject the thing.

JulienPalard:
@theacodes you mean to move the whole red ribbon on the cpython repo? Even its style?

theacodes:
Yeah I think so. The whole version switcher and concept of multiple release
versions is something unique to the cpython docs and isn't shared by other
users of this theme (like packaging.python.org and siblings). Putting it in
cpython will also make it more straightforward to translate with {% trans
%} directives.

I can be convinced otherwise though.

JulienPalard:
I think that's easier in cpython, also because not all versions are using the theme, so I'm moving it there. Thanks @theacodes.

theacodes:
Happy to take a look at the cpython change (even though I'm not a core dev,
I'm extremely familiar with the theme). Just @ me.

JulienPalard:
→ https://github.com/python/cpython/pull/13638

Issue #37, opened 2020-01-12
Release request


zware:
Hi, can we get a release of this with the latest fixes?  In particular, we're getting reports about the broken Sphinx link in the footer.

If there's anything I can do to help it along, please let me know.

JulienPalard:
I think I don't have the rights on pypi.org/p/python-docs-theme, cc @Mariatta @theacodes @ncoghlan?

Also @ewdurbin could we please have a "pre-prod" server for docs.python.org to test this (and to test docsbuild-scripts, new languages, ...) before pushing it to production?

Mariatta:
I have admin right. Let me know what I should be doing there...

Mariatta:
just publish this to pypi?

JulienPalard:
@Mariatta no no no! Publish this to pypi WHILE crossing fingers it won't break docs.python.org. I'll do too.

I'm really *not* a CSS guy, but I've read the 3 commits and it looks OK.

zware:
We'll need a version bump and tag before publishing; I can see to that if you want.

Mariatta:
😰been awhile since I have to publish to pypi manually using setuptools....

going through this checklist, and let me know if there are additional steps that I might miss
- [ ] bump version number?
- [ ] changelog?
- [ ] publish to test pypi?
- [ ] publish to real pypi?
- [ ] add .post1 to codebase?

Mariatta:
Here it is: https://pypi.org/project/python-docs-theme/2020.1/ 🤞

zware:
Thank you, @Mariatta!

PR #38, opened 2020-01-14
Bump version for release

zware:
For #37.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     name='python-docs-theme',
     # Version is date based as year.month[.serial], where serial is used
     # if multiple releases are needed to address build failures.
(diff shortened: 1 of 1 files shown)


Mariatta:
Do we have changelog to update?

Mariatta:
and thanks for creating the PR so quickly! I was about to do it

zware:
Doesn't appear that we do.

Mariatta:
Does anyone else have opinion on whether we should start adding changelog beginning at this release?

zware:
I think this project is low-traffic enough that interested parties can just read through the commit log.  This would only be the 21st commit in ~3 years, after all :)

Mariatta:
ok I'm good with that

PR #41, opened 2020-01-15
Remind user about virtualenv and using python -m

Mariatta:


freddrake:
@zware: You are certainly not alone.  I *never* activate virtual environments, and almost never interact with them directly.  Thankfully, we have tools that do these things for us.

Mariatta:
I'm not worried about advanced and experienced users who know what to do with venv.
This note is more for the newbies who may not have been initiated in the ways of virtual environments.

freddrake:
@Mariatta: I wasn't suggesting that you should hold off; given the current prevailing recommendations for workspace management, your change is the right thing.

Issue #30, opened 2019-07-05
Make the python documentation responsive for smaller viewports


karlcow:
Currently the theme is desktop only which makes the python documentation not very readable/accessible on mobile. 

There was an issue on the python project created 4 years ago to address this problem.
https://bugs.python.org/issue23312

Since then the documentation theme has been separated in a different package.

epicfaace:
Can we close this issue now? @JulienPalard

karlcow:
agreed.

PR #46, opened 2020-05-19
Make the theme responsive

obulat:
I have been reading on my phone a lot lately, and was disappointed by the fact that I couldn't read python docs: the fonts are too small, and, if I try to zoom in, the text doesn't fit on the page, so I have to swipe to see the whole line. 

This is my attempt to add responsiveness to python docs theme, fixing #30. 

On screen widths smaller than 900 px, it: 
- adds a mobile navigation bar with a 'hamburger' menu button, Python version switcher and a search box.

- adds a sliding menu that opens when 'hamburger' menu button is clicked. Menu contains language switching input and contents.

- removes the `related` bars

- removes the sidebar.

Screenshot with menu closed:
Responsive  Menu closed Galaxy S5
Screenshot with menu open:
Responsive  Menu open Galaxy S5

#### Problems that need to be solved:
- Normally, the mobile navigation bar should be placed in the `header` block of the layout, but python.org layout completely overrides it, so I added it in the body_tag block. To solve this, I would need to add changes to that code.

- I placed the switchers and other items on the menus as seemed logical to me, but feedback from the community is essential on this.

- I added a javascript file to open/close sidebar in a separate file. I could also add it into the layout file, but ideally all js files should be combined and minified.


python_docs_theme/layout.html
```diff
@@ -38,11 +38,71 @@
...
+        {% if not embedded %}
+            <script type="text/javascript" src="{{ pathto('_static/copybutton.js', 1) }}"></script>
+            <script type="text/javascript" src="{{ pathto('_static/menu.js', 1) }}"></script>
+        {% endif %}
     {% endif %}
     {{ super() }}
 {% endblock %}
 
+{%- block css %}
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
septatrix, line 50: This got added in sphinx 3.1 and should work mostly without problem for people using old sphinx versions. May however not hurt to include it for now and delete it latex

python_docs_theme/layout.html
```diff
@@ -38,11 +38,71 @@
...
+            {%- if pagename != "search" and builder != "singlehtml" %}
+                <div id="searchbox" style="display: none" role="search">
+                    <form class="search" action="{{ pathto('search') }}" method="get">
+                        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true"
+                             class="search-icon">
+                            <path fill-rule="nonzero"
+                                  d="M15.5 14h-.79l-.28-.27a6.5 6.5 0 001.48-5.34c-.47-2.78-2.79-5-5.59-5.34a6.505 6.505 0 00-7.27 7.27c.34 2.8 2.56 5.12 5.34 5.59a6.5 6.5 0 005.34-1.48l.27.28v.79l4.25 4.25c.41.41 1.08.41 1.49 0 .41-.41.41-1.08 0-1.49L15.5 14zm-6 0C7.01 14 5 11.99 5 9.5S7.01 5 9.5 5 14 7.01 14 9.5 11.99 14 9.5 14z"></path>
+                        </svg>
+                        <label id="mobilesearchlabel" for="searchbox">{{ _('Quick search') }}</label>
+                        <input type="text" name="q" aria-labelledby="searchlabel"/>
```
septatrix, line 75: The references on the labels do not work out. The `label` is for the `div` and the `input` is labelled by an element which does not exist. I would just put the input inside of the label and remove unneeded references/ids

python_docs_theme/layout.html
```diff
@@ -38,11 +38,71 @@
...
+                        <input type="submit" value="{{ _('Go') }}"/>
+                    </form>
+                </div>
+                <script>document.getElementById('searchbox').style.display='block';</script>
```
septatrix, line 79: Any specific reason to use this at all as `#searchbox` is already a block or not at least put it in the stylesheet

python_docs_theme/static/pydoctheme.css
```diff
@@ -1,11 +1,20 @@
 @import url("default.css");
 
+* {
+    box-sizing: border-box;
+}
```
septatrix, line 5: This should not be necessary and may break existing layouts

python_docs_theme/static/pydoctheme.css
```diff
@@ -17,7 +26,7 @@ div.related a:hover {
     color: #0095C4;
 }
 
-div.related:first-child {
+div.related:first-of-type {
```
septatrix, line 29: I have not seen any occurence where you shifted the `div.related` or is this due to some jinja block before which you added stuff?

python_docs_theme/static/pydoctheme.css
```diff
@@ -191,3 +200,228 @@ div.footer a:hover {
...
+        outline: 1px solid #999;
+    }
+    .nav-content .search label {
+        border: 0;
+        clip: rect(0 0 0 0);
+        height: 1px;
+        margin: -1px;
+        overflow: hidden;
+        padding: 0;
+        position: absolute !important;
```
septatrix, line 297: Please try to not use `!important`

python_docs_theme/static/pydoctheme.css
```diff
@@ -191,3 +200,228 @@ div.footer a:hover {
...
+        width: 1px;
+    }
```
septatrix, line 299: What exactly is the purpose of the label if you try to hide it (at least that what it seems so far from this code)

python_docs_theme/theme.conf
```diff
@@ -4,8 +4,8 @@ stylesheet = pydoctheme.css
 pygments_style = sphinx
 
 [options]
-bodyfont = 'Lucida Grande', Arial, sans-serif
-headfont = 'Lucida Grande', Arial, sans-serif
+bodyfont = 'SourceSansProRegular','Lucida Grande', Arial, sans-serif
+headfont = 'SourceSansProRegular','Lucida Grande', Arial, sans-serif
```
septatrix, line 8: This is unrelated and should be discussed in a seperate issue

python_docs_theme/static/menu.js
```diff
@@ -0,0 +1,22 @@
...
+        } else {
+            closeMenu();
+        }
+    });
+    doc.addEventListener('click', function () {
+        if (toggler.checked) {
+            closeMenu();
+        }
+    })
+})
```
septatrix, line 22: By moving the checkbox up in the dom tree it may be possible to implement this using only javascript and make it more accessible for people without javascript

python_docs_theme/layout.html
obulat: I added this to sphinx after creating this PR.

python_docs_theme/layout.html
obulat: I was trying to remove unnecessary use of jquery, as I read that this is planned in the future. However, I can make a PR later specifically addressing jquery removal.

python_docs_theme/static/pydoctheme.css
obulat: OK, I will remove this.

obulat:
Thank you for your review,  @septatrix !
I don't have a reliable internet connection at the moment, but will try to address your comments as soon as possible. Overall, I have tried to make the theme accessible as well as responsive, that's the reason for some label handling. Also, I tried to use only Vanilla javascript to make it easier to remove jquery dependency in the future. 

I would also like to have your feedback on the content for the mobile version. For example, the top navbar on mobile: Do you think logo, version switcher and searchbox are what should be there, or do we need to add/ remove something?

Also, the relbar is completely removed,  but I thought that it might be a good idea to add it at the bottom, without the searchbox, though.

python_docs_theme/static/pydoctheme.css
obulat: This label is there for accessibility reasons: screen readers can read it, but it doesn't crowd the layout. The search box is implemented this way on mdn website.

python_docs_theme/theme.conf
obulat: OK, I will add this as a separate PR for discussion. Python.org uses this font, so I changed it because I think SourceSans looks more modern.

python_docs_theme/layout.html
septatrix: I am all in for implementing this without using jquery but I do not see why this is necessary at all as a div is normally a block anyhow and if it is overwritten somewhere you could probably just change it in the stylesheet

python_docs_theme/static/pydoctheme.css
septatrix: Ah I see, so this is a manual version of `sr-only` like in bootstrap. I would probably just use a `aria-label="{{ _('Quick search') }}"` on the `input` element though

python_docs_theme/layout.html
obulat: I think I misunderstood your question.

I took the code for the search box from the current search box implementation and adapted it for mobile. There is a [jquery line $('.inline-search').show(0);](https://github.com/python/python-docs-theme/blob/628a66466589ba42037d4aad3d676cb7d81fb6e6/python_docs_theme/layout.html#L23) which I converted to vanilla JS, both in the mobile part and the full version. I thought you were asking why I converted it to JS the way I did.

Are you, instead, asking why that line is there at all? If so, I don't know the answer :)

septatrix: Yes that is what I was referring to. I have not tried this branch yet so I do not know if removing it makes a difference though what confuses me is that the style in the stylesheet make the element an `inline-block` by default. Maybe there is a special reason for it... :shrug:

python_docs_theme/static/pydoctheme.css
obulat: Yes, I've added mobile menu before div.related, because I had to add it to the top of the `body` block.

python_docs_theme/layout.html
obulat: I simplified the labels.

python_docs_theme/layout.html
obulat: I removed these lines, and didn't notice any change...

python_docs_theme/static/pydoctheme.css
obulat: Replaced visually-hidden label for accessibility by `aria-label`s

python_docs_theme/static/menu.js
obulat: I replaced the checkbox with a button. Is this better for people without javascript?

JulienPalard:
Beware, i'm working on moving the switchers generation code to docsvuild scripts instead of cpython. See https://github.com/python/docsbuild-scripts/issues/90

I did not think about switchers being located in different places, we have to check if there is something to fix either in docsbuild side or here.

As the switchers are no longer to be maintained in cpython, I was removing the placeholders and the swotchers variable in https://github.com/python/cpython/pull/20969

I'm on my phone right know so I can't test it, but we need to agree on how to "communicate" to docsbuild the switchers places.

python_docs_theme/static/menu.js
septatrix: What I would have done is move the input as a checkbox directly next to the sidebar, reference it in the label via id and apply the styles based on whether the checkbox is checked or not. I may have time tomorrow to create a PR on your fork repo if you want to.

obulat: I have done it like this at first (checkbox input and applying styles based on :checked property). Then I was checking for accessibility and found [this article](http://www.ashleysheridan.co.uk/blog/Making+an+Accessible+Hamburger+Menu) which shows that a button as menu opener has some advantages regarding accessibility: namely, it is easier to set up keyboard handling. On the other hand, I just realized that people probably don't use keyboard for navigation on mobile, and it is more important to set up No-script solution than keyboard navigation on mobile. 
Also, I looked at other documentation sites: readthedocs theme and VueJS docs, and their hamburger menu doesn't work without javascript.

JulienPalard:
Hi @obulat!

Took the time to think about it, but beware, I'm particularily bad at HTML/JS:

- python-docs-theme is not cpython specific, it can be used elsewhere, that's typically why the "This document is for an old version of Python" header [is in cpython](https://github.com/python/cpython/blob/master/Doc/tools/templates/layout.html) and not in the theme.
- Switchers are *not* cpython specific: other documentations can be translated and have versions.
- The classes `.version_switcher_placeholder` and `.language_switcher_placeholder` has always been in use and looks good to me.
- Having multiple version placeholders may work, IIRC, jquery will properly create them all when I do `$('.version_switcher_placeholder').html(version_select);`

The current `switchers.js`, [the one from docsbuild-scripts](https://github.com/python/docsbuild-scripts/blob/master/templates/switchers.js) has two ways of working (see function create_placeholders_if_missing):

- Either there's a '.version_switcher_placeholder' and it uses it (typically when cpython or the theme places it)
- Either there's not, and it tries it best to find a place to create both placeholders (side by side), see the ugly `probable_places` array (I may soon add a cleaner [cpython-language-and-version](https://github.com/python/cpython/pull/20969/files#diff-dc15d4d8d8fb2a558300077d17dd1b21R15) in front of it, but it may not be usefull if you handle it in the theme).

What about adding all needed placeholders in python-docs-theme, so `switchers.js` from docsbuild-scripts could rely on them, instead of trying to inject HTML? In "mobile version" it currently injects them in HTML that is not rendered, so no switchers appear.

How to test with docsbuild-script locally:

```bash
$ # In a clone of docsbuild-scripts
$ mkdir -p www logs build_root
$ python3 -m venv build_root/venv/
$ build_root/venv/bin/python -m pip install -r requirements.txt
$ build_root/venv/bin/python -m pip install path_to_your_python-docs-theme
$ python3 ./build_docs.py --quick --build-root build_root --www-root www --log-directory logs --group $(id -g) --skip-cache-invalidation --language en --branch master
```

obulat:
Hi, @Julien, thank you for your comments, that's really helpful!

I will try to look into it as soon as I have time.

python_docs_theme/static/menu.js
septatrix: In addition if you follow [this guide](https://www.silvestar.codes/articles/css-sidebar-toggle/) you also get a working css only solution with added accesibility using javascript

septatrix:
I think a problem which should be tackled first is overflowing stuff like `table`s, all `dt` for definitions as well as `code` segments inside the docs. These are for the most part not broken an lead to a wider page than intended. Together with the addition of the `<meta name="viewport" ...>` tag this will already solve most problems. This way no major design changes would be necessary as the sidebar should actually also work great on mobile devices (maybe make it a few pixels wider). The problem with the header/footer wrapping badly due to its floating nature already happens for me at around 1200px width.

obulat:
> I think a problem which should be tackled first is overflowing stuff like `table`s, all `dt` for definitions as well as `code` segments inside the docs. These are for the most part not broken an lead to a wider page than intended. Together with the addition of the `<meta name="viewport" ...>` tag this will already solve most problems. This way no major design changes would be necessary as the sidebar should actually also work great on mobile devices (maybe make it a few pixels wider). The problem with the header/footer wrapping badly due to its floating nature already happens for me at around 1200px width.

I've added wrapping `tables` in a `div` that can be horizontally scrolled: this is the solution used in readthedocs. `dt` and `code` segments are already horizontally scrollable.

python_docs_theme/static/pydoctheme.css
```diff
@@ -191,3 +196,254 @@ div.footer a:hover {
...
+        min-width: 100%;
+        padding: 0;
+        font-size: 0.875rem;
+    }
+    div.bodywrapper {
+        margin: 0;
+    }
+    /* Typography */
+    div.body h1, div.body h2, div.body h3, div.body h4, div.body h5, div.body p {
+        font-family: SourceSansProRegular, Arial, sans-serif;
```
septatrix, line 212: Why do you change the font family for a specific width?

python_docs_theme/static/pydoctheme.css
```diff
@@ -191,3 +196,254 @@ div.footer a:hover {
...
+        width: 100%;
+        position: fixed;
+        top: 0;
+        left: 0;
+        background-color: white;
+        box-shadow: rgba(0, 0, 0, 0.25) 0 0 2px 0;
+        z-index: 1;
+    }
+    .mobile-nav * {
+        box-sizing: border-box;
```
septatrix, line 245: If possibly I would stay with the sphinx default of not modifying the box-sizing property

pradyunsg:
Wheee! This is amazing. Is there anything I can do to help move this forward? :)

obulat:
Hi, should I close this issue due to the new theme added in https://github.com/pypa/pip/pull/9012 ?

pradyunsg:
@obulat No no. I don’t imagine docs.python.org would be changing to that theme anytime soon, and this PR is a pretty great improvement over status quo.

septatrix:
And even if the theme were to change there may still be other projects using this theme

python_docs_theme/static/pydoctheme.css
obulat: I tried removing it, but I couldn't get the searchbox to fit the screen on all the different sizes: flexbox and padding and border in `content-box` sizing model are too complex. I don't think adding a different box-sizing property should be a problem because the new box-sizing applies only to the `.mobile-nav` part.

obulat:
> What about adding all needed placeholders in python-docs-theme, so `switchers.js` from docsbuild-scripts could rely on them, instead of trying to inject HTML? In "mobile version" it currently injects them in HTML that is not rendered, so no switchers appear.
> 

Hi, @JulienPalard , 

I'm on Windows, so originally I couldn't test the `docsbuild-scripts` steps because they wouldn't run. Now, I tried running them on my Ubuntu (Windows Subsystem for Linux), but I couldn't get my local version of the theme to apply. 

However, I did add empty placeholder divs: `.version_switcher_placeholder` in the top navbar, and `.language_switcher_placeholder` in the left menu.
Could you please test it?

python_docs_theme/static/pydoctheme.css
obulat: I removed this change.

obulat:
I'm sorry it took me so long to get back to implementing the requested changes. 
@pradyunsg , thank you for your comments!

The index page still doesn't look too good, with the main reason being that the layout used for contents is a table, which doesn't look good on a small screen. I think we should convert it into a two-column layout that can be converted into a one-column layout on smaller screens. However, I would say it is a to-do for another PR.

JulienPalard:
> I'm on Windows, so originally I couldn't test the `docsbuild-scripts` steps because they wouldn't run.

There's no Windows support on docsbuild-scripts, and I don't except to add it anytime.

> Now, I tried running them on my Ubuntu (Windows Subsystem for Linux), but I couldn't get my local version of the theme to apply.

This is really strange, can you open an issue on docsbuild-script with the steps you followed so I can try reproduce and we fix this? I often build the docs locally (on Debian), and have no theme issues.

JulienPalard:
@obulat I was able to build a local version, I had to patch docsbuild-scripts (hardening a regex detecting the version part in the URL in case the URL contains an IP like 0.0.0.0 ...).

It works: it gets replaced \o/ but only on mobile, as the placeholder only exist for mobile.

obulat:
> There's no Windows support on docsbuild-scripts, and I don't except to add it anytime.

I completely understand, just wanted to add this bit of information for context.

> This is really strange, can you open an issue on docsbuild-script with the steps you followed so I can try reproduce and we fix this? I often build the docs locally (on Debian), and have no theme issues.

I will try to do that as soon as I can.

obulat:
> @obulat I was able to build a local version, I had to patch docsbuild-scripts (hardening a regex detecting the version part in the URL in case the URL contains an IP like 0.0.0.0 ...).
> 
> It works: it gets replaced \o/ but only on mobile, as the placeholder only exist for mobile.

As this PR is for the mobile version, I did not touch the placeholders for the desktop version. I could add them here, or in a new PR.

JulienPalard:
> As this PR is for the mobile version, I did not touch the placeholders for the desktop version. I could add them here, or in a new PR.

I can understand, but this is a bit dangerous: if there's a release between your two PRs, language and version switchers will disaperar on desktop (as paceholders are found on the new mobile part, my code don't try to add them on the desktop part).

I mean, in switchers.js I have two branches : either the placeholders are found and I use them, or they are not found and I create them, so if they're found, they're not created.

obulat:
> I can understand, but this is a bit dangerous: if there's a release between your two PRs, language and version switchers will disappear on desktop (as placeholders are found on the new mobile part, my code don't try to add them on the desktop part).

I didn't realize that. I tend to include too much in my PRs, so I was trying to improve myself 😄 

I added the placeholders.

obulat:
Hi, @JulienPalard , @septatrix , sorry for pinging you. Are there any updates on this PR?

JulienPalard:
I just updated https://github.com/python/docsbuild-scripts/ so we can test language and version switchers in themes PR faster!

It goes like this now:

```
python3 ./build_docs.py --quick --build-root ./build_root --www-root ./www --log-directory ./logs --group $(id -g) --skip-cache-invalidation --theme 'git+https://github.com/obulat/python-docs-theme@master' --language en --branch 3.9
```

but I do hit a:

```
Theme error:
An error happened in rendering the page about.
Reason: TemplateAssertionError("block 'rootrellink' defined twice")
```

I did not diagnosed why yet (free time is melting like icebergs).

obulat:
Thank you for letting me know, Julien!
> 
> ```
> Theme error:
> An error happened in rendering the page about.
> Reason: TemplateAssertionError("block 'rootrellink' defined twice")
> ```
> 
I do see that I add the definition of the rootrellink block, so the page probably has two definitions of the same block. I will have to look at it and find a way to fix it in my free time :)

obulat:
After some review, I found the reason for the error:

I've copied the `relbar` macro from the `sphinx/themes/basic/layout.html` to add the `language_switcher_placeholder` and `version_switcher_placeholder` divs to it. What I didn't realize that the `rootrellink` block definition inside the `relbar` macro had been overridden in the `layout.html` before, so there were 2 definitions of `rootrellink` block.

In the new commit, I moved the python docs theme's  `rootrellink` declaration into the `relbar` macro, so there are no more duplicates. Could you test if it works when you have free time, @JulienPalard ?

JulienPalard:
@obulat Took the time to test, I'm not happy when testing something is not easy... I think we have to enhance this...

I'm getting:
```
Theme error:
An error happened in rendering the page about.
Reason: TemplateAssertionError("block 'relbaritems' defined twice
```

obulat:
@JulienPalard Thank you for testing!  I guess it's the same error again, I wish I could have checked it. I will try to fix it today or tomorrow.

obulat:
Hi, @JulienPalard, I got it to work! I removed the duplicates, adjusted styles a little bit, and uploaded the built docs to github pages: https://obulat.github.io/test_docs/

python_docs_4

Do we have a better-quality Python logo?

pradyunsg:
https://www.python.org/community/logos/ maybe?

https://logos-download.com/9988-python-logo-download.html seems to have an SVG version of just-the-logo as well, which thanks to the magic of vector graphics, should fit in wherever we want.

JulienPalard:
I tested it, and found some issues:

- I have the version switcher in a "narrow window", but it disapear in a "wide window".
- I don't see the language switcher.
- I have an unclickable version switcher when I split my screen in two (cf. screenshot 1)
- When, in "full screen" I:
    - Fold the menu
    - Refresh
    - Unfold the menu
    - Split my screen vertically (I'm using a window manager)
    - I have a stange big white left margin (cf. screenshot 2)

<details>
<summary>
Screenshot 1
</summary>

Screenshot_2021-04-13 The Python Tutorial — Python 3 9 4 documentation

</details>

<details>
<summary>
Screenshot 2
</summary>

Screenshot_2021-04-13 The Python Tutorial — Python 3 9 4 documentation(1)

</details>


Now that the theme is responsive, I don't think we should keep the "autofolding menu cookie", I'm not happy with the flickery loading of it, I think we do loose some SERP positions due to search engines not liking this flickery loading, and it may be a bug magnet (4th issue / screenshot 2). Do someone actually use this folding mecanism? I never used it myself.

JulienPalard:
@obulat Motivated by your work on the theme, I'm trying to auto-build the HTML on PR's, I don't know if you'll have to rebase on top of 6d56e32 or not, but if I did it properly, when you push, we should all have a build artifact containing the html files for local viewing.

obulat:
> I tested it, and found some issues:

I have seen the problem with the version/language switchers not being visible, or having a width of 0 or 1px.
When I was testing this, I had to delete the `docs-build-scripts` and clone them again to see the changes on this branch: the scripts somehow cache the `https://github.com/obulat/python-docs-theme` code and I don't know how to make them re-download it.

python_docs_theme/static/pydoctheme.css
pradyunsg: FWIW, I don't see any reason to avoid doing this, as long as whoever works on this theme finds that model easier to work with. There are variants of normalise.css that set this on the entire page, so setting this for a whole bunch of elements, is A-OK in my opinion.

JulienPalard:
> I had to delete the `docs-build-scripts`

You can `rm -fr build_root` inside `docs-build-scripts` instead of dropping docs-build-scripts entierly:
- it drops the venv, so it properly reinstall your theme
- it drops sphinx build and cache properly regenerating all the pages with the new theme

should be enough.

willingc:
Hi @obulat, Thanks for the work that you have been thoughtfully doing on this PR. We're starting up a Documentation Workgroup (we'll announce the application process in about a month) and a Documentation community for official Python docs. I would love to see you involved if you are interested. One of the key items that I would like to take on this year and bring to fruition is a responsive, dark theme with better typography. I will give this a closer look in the coming week. 

@pradyunsg, if you have time can you see what the differences are between this PR and the Furo theme.

@obulat, if you have time can you also compare the work on this PR with the [Furo theme](https://github.com/pradyunsg/furo). Likes, areas of improvement.

@JulienPalard Do we have any sense of how many folks are using this theme other than CPython?

Thanks all!

cc/ @Mariatta

Mariatta:
I know that pylint uses python-docs-theme: http://pylint.pycqa.org/en/latest/
Looking at GitHub's "Dependencies", it looks like there are 315 repos uses this, though most looks like forks 

https://github.com/python/python-docs-theme/network/dependents

pradyunsg:
> @pradyunsg, if you have time can you see what the differences are between this PR and the Furo theme.

There are lots! Outside of the structural/content-styling differences though, this PR looks great for improving the not-great UX of the Python Docs on mobile.

For anyone who wants to see these docs in action, @obulat hosted them here: https://obulat.github.io/test_docs/

pradyunsg:
> Do we have any sense of how many folks are using this theme other than CPython?

- Sphinx documentation on GitHub using this theme (gives 338 results right now): https://github.com/search?q=python_docs_theme+filename%3Aconf.py&type=Code
- Sphinx Themes on GitHub that inherit from `python_docs_theme` (just pypa-docs-theme): https://github.com/search?q=python_docs_theme+filename%3Atheme.conf&type=Code
- PyPI packages that declare a dependency on this theme (four, one of them being PyLint): https://libraries.io/pypi/python-docs-theme/dependents

I think that covers nearly all easily-discoverable users?

obulat:
> Hi @obulat, Thanks for the work that you have been thoughtfully doing on this PR. We're starting up a Documentation Workgroup (we'll announce the application process in about a month) and a Documentation community for official Python docs. I would love to see you involved if you are interested. One of the key items that I would like to take on this year and bring to fruition is a responsive, dark theme with better typography. I will give this a closer look in the coming week.

Thank you, @willingc, I would be really happy to join the Workgroup, and help improve the way Python documentation looks.

> @obulat, if you have time can you also compare the work on this PR with the [Furo theme](https://github.com/pradyunsg/furo). Likes, areas of improvement. This is the general [project tracking issue on the Docs Community repo](https://github.com/python/docs-community/issues/1). ☀️

Furo has a more modern look, has a dark mode, a better experience when copying code, and better spacing between clickable elements, which is important for SEO and User Experience. I also like that the Contents/navigation around the site, including the breadcrumbs in the footer, are more semantically correct there. Also, a larger line-height makes reading experience easier.

obulat:
> I tested it, and found some issues:
> 
> * I have the version switcher in a "narrow window", but it disapear in a "wide window".

Could this bug be due to the previous version of the theme being loaded somehow? I have seen the same bug before, and fixed it, and now the[ preview site ](https://obulat.github.io/test_docs/) doesn't have this problem.

> * I don't see the language switcher.
The language switcher should be in the sidebar in the mobile version.

> * I have an unclickable version switcher when I split my screen in two (cf. screenshot 1)

> * When, in "full screen" I:
>   
>   * Fold the menu
>   * Refresh
>   * Unfold the menu
>   * Split my screen vertically (I'm using a window manager)
>   * I have a stange big white left margin (cf. screenshot 2)

Could these problems be also due to the previous version of the theme? I don't know how to split my screen and to test it.

> Now that the theme is responsive, I don't think we should keep the "autofolding menu cookie", I'm not happy with the flickery loading of it, I think we do loose some SERP positions due to search engines not liking this flickery loading, and it may be a bug magnet (4th issue / screenshot 2). Do someone actually use this folding mecanism? I never used it myself.

Should I try to remove it? I never fold the side menu...

pradyunsg:
> Should I try to remove it? I never fold the side menu...

Can this be a follow up PR? We don't have to get this perfect in the first shot, right?

JulienPalard:
> > Should I try to remove it? I never fold the side menu...
> 
> I feel this can this be done in a follow up PR, if we want to do it. We don't have to get this perfect in the first shot, right?


I agree with @pradyunsg, it's for another PR / another discussion.

JulienPalard:
@obulat Could you rebase on top of master? I would like to see if https://github.com/python/python-docs-theme/pull/69 works.

JulienPalard:
In the preview site everything looks find, but when I try it locally using:

    $ rm -fr www/ build_root/venv-*/ build_root/3.9/cpython-en/Doc/build/; python3 ./build_docs.py --quick --build-root ./build_root --www-root ./www --log-directory ./logs --group $(id -g) --skip-cache-invalidation --theme 'git+https://github.com/obulat/python-docs-theme@master' --language en --branch 3.9

Here's what I'm seeing:
- 1116×379: menu is expanded, no language switcher, no version switcher.
- 379×1116: Everything is fine.
- Desktop mode: No language switcher, no version switcher.

If there's anything I can do to enhance those switchers situation, don't hesitate to tell me, I agree that the current situation is far from confortable... the injection of a script tag using sed right before the `</body>` is ... #sry.

One idea I have in mind since [a long time](https://mail.python.org/archives/list/python-ideas@python.org/message/Y35KK55BW7PTUIQKSNROLZGYVHBMHE7O/):
- The server could put a `versions.js` file at the root of the doc.
- The theme could try to fetch it, and if it exists build switchers according to the described content.

(Sadly it would probably also mean a redraw of the page after load, which is bad for SEO?)

This way the theme will still be compatible with other project without version/translations, they could opt-in for the switchers by placing a simple file, ...

Another idea would be to study how other do it (readthedocs / sphinx / ...).

JulienPalard:
@obulat Yeah, the github build works! It would, I hope, ease the review of PRs a lot.

I'm seeing the same behavior on my local build and on the github build: no switchers on desktop mode, what do you think?

obulat:
The GitHub actions build is awesome! I can see what is built right away! The only problem is that when the build fails, the error message shown is just 'Cannot find sitemap.xml', which is not the actual error in the build.

The problem is caused by the fact that I put the switchers between two `li` items ( I don't know why I did that, I guess I didn't quite understand how the templates work when I did that). Those divs are not added to the build. 

On the GitHub pages, I have `span`s with `version_switcher_placeholder` and `language_switcher_placeholder` in the top menubar (relbar), which are put there by the `switchers.js`.

obulat:
I've fixed the switcher placement in the desktop mode, here is the link to the GitHub action run page with the build artifact: https://github.com/obulat/python-docs-theme/actions/runs/804990741 .

JulienPalard:
> The only problem is that when the build fails, the error message shown is just 'Cannot find sitemap.xml', which is not the actual error in the build.

I'm in the process of fixing this, thanks for reporting it: https://github.com/python/docsbuild-scripts/pull/104

JulienPalard:
Looks good to me!

JulienPalard:
Hum, when I'm on `http://0.0.0.0:8000/3.9/contents.html` I find no way to get back to `http://0.0.0.0:8000/3.9/`.

obulat:
If you click on `3.9.4 Documentation` link to the right of the switchers, you should be able to get to the `/3.9` page.
By the way, the version number on that link is probably redundant: it is repeated twice on the same line: 

Screen Shot 2021-05-03 at 11 06 54 AM



> http://0.0.0.0:8000/3.9/contents.html

How do you normally open the 'contents' page? I can open it at this URL, but I don't know how to get to it from the homepage...

pradyunsg:
I think it's about not being able to reach the home page from the mobile view?

It is a valid concern and something worth fixing. OTOH, I also feel like, it's fine to address this in a follow up too. :)

obulat:
> I think it's about not being able to reach the home page from the mobile view? That hides "Python" IIUC.
> 
> It is a valid concern and something worth fixing.

Oh, I see. Definitely a valid concern. Currently, in the mobile view there is a python logo which has no link. I can add a link to python.org to it, just like in the desktop version, and add another link to the documentation home. However, this is a bad decision because it doesn't fit well in English, I'm sure it will spill over in other languages. And the search input becomes very small.
Untitled(1)

Another option is link the logo to the documentation home (http://0.0.0.0:8000/3.9/). A link to python.org can be added either in the sidebar or in the footer.

--------
### This is what I would prefer to do in the mobile view:
1. Add a link to documentation home to the python logo
2. Add a link to python.org at the top of the sidebar.

But I am open to any other suggestions!
 
 P.S. The screenshot shows that we need to increase links' line height, and decrease the left padding for lists, at least in the mobile view. It can be done in a different PR, though :)

JulienPalard:
Don't wait for me to give UI/UX advices, my skill stop at testig (I see how bad the website I do myself look), I trust you to make the good choices.

pradyunsg:
The plan sounds good to me! And, as I'm probably over-stating now, we can iterate too! :)

python_docs_theme/layout.html
```diff
@@ -85,7 +85,10 @@ <h3>{{ _('Navigation') }}</h3>
         <span></span>
     </label>
     <nav class="nav-content" role="navigation">
-        <img src="{{ pathto('_static/' + 'py.svg', 1) }}" alt="Logo"/>
+         <a href="{{ pathto('index') }}">
```
pradyunsg, line 88: pathto(master_doc) would be more appropriate here?

pradyunsg: Or does CPython documentation use contents for the toctree?

obulat: Thank you for such a quick review! :) I know what I added doesn't work for sure. Let me try pathto(master_doc).

obulat: I need help! 
I don't know how to add the correct link to the `docs.python.org/3.9/`. 
In line 88 in layout.html I added `<a href="{{ pathto(theme_root_url) }}" `, and it creates `file:///.../doc-html(3)/3.9/https://www.python.org/.html`

pradyunsg: Don't wrap it in pathto, just `{{ theme_root_url }}` is enough.

That should fix it. :)

obulat: Hmmm. It now links to python.org...

pradyunsg: That's what "root_url" in theme_config_options is set to in the conf.py file of the documentation being built.

pradyunsg: In other words, that is correct and I thought that's what you wanted?

obulat: I guess I was confusing the two kind of links (to the docs home and the language/framework home). On the sites I used a lot lately (eg Vue.js) the style of the framework home page and the style of the documentation homepage are the same. And it feels like I just go to the documentation home, not the general home, when I click on the logo.

But you are right, now we have the correct behavior, linking to python.org.

We still don't have a way of going to the documentation homepage, `docs.python.org/3` from other pages on mobile. Is it necessary to add it, and add it in this PR specifically?

pradyunsg: If you want to have a link to the documentation homepage, that'd be `{{pathto(master_doc)}}` or `{{ pathto("index") }}`. No need for a |e on either of them.

pradyunsg: Honestly, this feels like something we can do in a follow up PR.

I'd really like to get this landed soon, so that we can start moving other things forward without needing to come back and update this.

obulat: Please, let me know if I need to change anything? I would be happy to get this landed soon, too!

JulienPalard:
LGTM. @pradyunsg if it's OK for you too, I'm ok for a merge :)

willingc:
Way to go @obulat :tada: 🐍!  Thanks for the great teamwork @JulienPalard and @pradyunsg. Happy to see this merged soon 👍🏼

willingc (approved):

JulienPalard (approved):

obulat:
Thank you everyone! I'm really excited for this change!

pradyunsg (approved):
🚀

FGuillet (approved):

JulienPalard:
Thanks **a lot** @obulat, I hope this can go to production soon. @willingc do you know the release process?

I don't have the upload rights to pypi/p/python-docs-theme.

willingc:
I don't @JulienPalard, but I can look into it tomorrow.

birkenfeld:
Thanks also from me - when I originally put together this theme I hadn't even expected it to live so long :D

willingc:
@mariatta @ncoghlan @theacodes If one of you can deploy this awesome update or give @JulienPalard or me privileges that would be great. Would love to see @obulat's work in production :tada:

@birkenfeld Thank you for all that you have done for docs over the years. You rock. :tada:

theacodes:
@willingc you now have permissions on PyPI ✨

septatrix:
If there are already so many people taking a look at this right now I would like to mention the dark theme PR which also is ready for review

willingc:
@septatrix I just saw this now. After the dark theme lands we can do a dot release to get that in.

PR #54, opened 2020-09-16
Fix hyphenation in Chrome

hugovk:
There's a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1029214) or [two](https://bugs.chromium.org/p/chromium/issues/detail?id=1022011) in Chromium (e.g. Chrome, Opera) such that the Lucida Grande font doesn't hyphenate properly (with `hyphens: auto`) which makes things harder to read.

This shows up on macOS, which has this this font, but not on Android or Windows, which use the fallback Arial. Firefox and Safari are okay.

This affects at least [docs.python.org](https://docs.python.org/3.10/tutorial/index.html) and [pip.pypa.io](https://pip.pypa.io/en/stable/development/release-process/), and both use this theme.

# Before

Here's an example from https://pip.pypa.io/en/stable/development/release-process/ in newest Chrome 85 on macOS Mojave:

docs1

*(Created with `sphinx-quickstart` using the defaults, changing `html_theme = 'alabaster'` to `'python_docs_theme'`, putting some pip text in `index.rst` and running `rm -rf _build; make html`: [sphinx-test2.zip](https://github.com/python/python-docs-theme/files/5230007/sphinx-test2.zip).)*

# After

This PR suggests to ditch Lucida Grande, at least until when and if the Chrome bug is fixed and widely deployed. This would also make it consistently use Arial as on non-Macs:

docs2

---

See also my original PR at https://github.com/pypa/pip/pull/8879 to disable `hyphens: auto`, but they'd prefer to fix the upstream theme.

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -57,7 +57,7 @@ div.sphinxsidebar a:hover {
 
 form.inline-search input,
 div.sphinxsidebar input {
-    font-family: 'Lucida Grande',Arial,sans-serif;
+    font-family: Arial,sans-serif;
     border: 1px solid #999999;
     font-size: smaller;
     border-radius: 3px;
--- a/python_docs_theme/theme.conf
+++ b/python_docs_theme/theme.conf
@@ -4,8 +4,8 @@ stylesheet = pydoctheme.css
 pygments_style = sphinx
 
 [options]
-bodyfont = 'Lucida Grande', Arial, sans-serif
-headfont = 'Lucida Grande', Arial, sans-serif
+bodyfont = Arial, sans-serif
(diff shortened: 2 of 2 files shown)


hugovk:
Closing this, Chrome has now fixed this in Canary 88.0.4300.0!

I've confirmed it working for https://pip.pypa.io/en/stable/development/release-process/ in Canary Version 88.0.4301.0 (Official Build) canary (x86_64) on macOS (and still repro on older Chrome 86).

Chrome 88 is due out on 2021-01-19: https://www.chromestatus.com/features/schedule

JulienPalard:
Did not commented before, because I have literally no front-end skills, but I'm relieved that we don't have to fix Google bugs our side, it felt wrong.

PR #67, opened 2021-03-07
Updated copyright link to license in repo

jablonskidev:
This should fix issues #652 and #653 in the devguide. (The issues are duplicates of each other, so this PR is still addressing only one issue at a time.)


python_docs_theme/layout.html
```diff
@@ -45,7 +45,7 @@
 
 {% block footer %}
     <div class="footer">
-    &copy; <a href="{{ pathto('copyright') }}">{% trans %}Copyright{% endtrans %}</a> {{ copyright|e }}.
+    &copy; <a href="https://github.com/python/devguide/blob/master/LICENSE">{% trans %}Copyright{% endtrans %}</a> {{ copyright|e }}.
```
Mariatta, line 48: The python-docs-theme is a common library, and used not just by devguide but also other projects, like [Python Packaging](https://packaging.python.org/) and [pylint](http://pylint.pycqa.org/en/latest/).
 
So perhaps we need a way to customize this.

CAM-Gerlach: Random contributor stopping by, but what about a theme variable (with the default being a URL to the canonical PSF license)?

Mariatta: Having a theme variable sounds great, so it can be configured for each project.

jablonskidev: Ah yes, that would make much more sense! @CAM-Gerlach Thanks for the suggestion. @Mariatta Thanks for the feedback. @CAM-Gerlach I'm less familiar with Sphinx themes, so I likely won't have a solution right away. If you'd like to propose specific changes, then you are more than welcome to.

CAM-Gerlach:
```suggestion
    &copy; <a href="{{ theme_copyright_url }}">{% trans %}Copyright{% endtrans %}</a> {{ copyright|e }}.
```

Sure, you'd just apply the above change and add the following line to the `theme.conf` (using the current de-facto value as the default, for backwards compat):

```ini
copyright_url = https://docs.python.org/copyright.html
```

And then any theme user that wanted to customize the link could do so in the `html_theme_options` dict in the `conf.py`, i.e.

```python
html_theme_options = {
    "copyright_url": "https://www.example.com",
}
```

Lmk if I can be of any further assistance!

CAM-Gerlach: [Here's](https://github.com/spyder-ide/spyder-docs-sphinx-theme/commit/5f9fe5d90a16b8b15261ef4eacd53a9040fd539e) a relatively minimal example of adding a footer link, if that's helpful.

CAM-Gerlach:
```suggestion
    &copy; {% if theme_copyright_url or hasdoc('copyright') %}<a href="{% if theme_copyright_url %}{{ theme_copyright_url }}{% else %}{{ pathto('copyright') }}{% endif %}">{% endif %}{% trans %}Copyright{% endtrans %}{% if theme_copyright_url or hasdoc('copyright') %}</a>{% endif %} {{ copyright|e }}.
```

Here's a better solution that is both more backwards-compatible, more functional and does the "right thing" more of the time:

* If the new `copyright_url` theme option is defined, uses that,
* Otherwise, if a file named `copyright` exists in the repo, links to that (current behavior, fully backward compatible)
* Otherwise, doesn't make copyright a link at all, so it isn't broken

The would be coupled with the following in the `theme.conf`:

```ini
copyright_url =
```

This would be fully backwards-compatible and make more sense especially for non-PSF projects, so they wouldn't point to your license by default that doesn't apply to them.

jablonskidev: @CAM-Gerlach  Thanks for taking the time to explain this. Since you kindly asked if you could be of more assistance, there is something I'd like to understand more clearly.

Using a theme variable as you suggested sounds like the best option here, but I'm a little fuzzy on how to leave options open for all users of the python-docs-theme library while also setting a specific option for when Python's devguide is using that library:

- If I define a `copyright_url` theme option in `theme.conf`, then I'm making changes to the library rather than to how the devguide is using the library, so that doesn't seem like a viable option.

- If I create a `copyright` file in the python-docs-theme library, then I have the same problem. I wondered if you might have meant creating a `copyright.rst` in the devguide repo, but that would mean having the same information in multiple places, which isn't ideal.

As a result, the cleanest solution seems to me to be the first one you suggested, where you:

- Add a theme variable to `layout.html` with no conditionals
- Add a default URL to `theme.conf`
- Specify which URL the devguide will be using for the copyright by adding a key-value pair to the dictionary `html_theme_options` inside the devguide's `conf.py`

As I type all this it, it occurs to me that maybe you were proposing that we add a theme variable with the conditionals you suggested *and* add the URL to the key-value pair to the dictionary `html_theme_options` inside the devguide's `conf.py`. Is that what you meant?

CAM-Gerlach: > As I type all this out, it occurs to me that maybe you were proposing that we add a theme variable with the conditionals you suggested _and_ add the URL to the key-value pair to the dictionary `html_theme_options` inside the devguide's `conf.py`. Is that what you meant?

Yup, exactly; I just didn't explicitly specify that again, sorry. My revised proposal is essentially a superset of the original; it still allows consumers to set a URL (or local path) of their choice in their `conf.py` just like the previous (i.e. if `copyright_url` is specified in `conf.py` theme options, the behavior is identical), while behaving much better if it is absent:

* Keeping the current behavior by default if `copyright.rst` is present, so themes can still use their existing local `copyright.rst` file if they prefer, without having explicitly change their config and hardcode the path
* In case that file is missing and no link is specified, don't generate a broken link (to fix the proximate bug in python/devguide#652 and python/devguide#653 for all theme users automatically, without any further user action)
* Not automatically changing the URL for all existing users to point to Python's own copyright statement, unless they take explicit action to specify their own (avoiding any unintended legal uncertainty for either users or the PSF)

This does mean PSF projects that don't include their own `copyright.rst` and want to point to the standard Python copyright statement will need to add one line to their `html_theme_options` with the link, but I would think that relatively small one-time cost is less than that of affecting backward compat and defaulting to a potentially legally problematic value for non-PSF users, and still fixes the immediate bug (broken link) without users having to do anything.

jablonskidev:
@CAM-Gerlach Thanks again, that was super helpful, and I learned more about Sphinx along the way. I committed your second suggestion and will also submit a PR to add the license URL to the dictionary `html_theme_options` inside the devguide's `conf.py`.

CAM-Gerlach:
You're welcome, I'm glad! I'm no expert at Sphinx theming either; I know just enough to be dangerous :laughing: Looks like its up to @Mariatta for the final review?

matrixise:
Could you explain how to use it? because with my workflow, it doesn't work.

```
git clone https://github.com/python/devguide
cd devguide
make venv
./venv/bin/pip install -U pip
./venv/bin/pip uninstall python-docs-theme
./venv/bin/pip install git+https://github.com/jablonskidev/python-docs-theme.git@patch-1
```

patch with this diff
```diff
diff --git a/conf.py b/conf.py
index 4071095..f20b61f 100644
--- a/conf.py
+++ b/conf.py
@@ -95,6 +95,7 @@ pygments_style = 'sphinx'
 # Use the upstream python-docs-theme
 html_theme = 'python_docs_theme'
 html_theme_options = {
+    'copyright_url': 'https://www.demo.org',
     'collapsiblesidebar': True,
     'issues_url': 'https://github.com/python/devguide/issues/new',
 }
```

```
> make html
./venv/bin/sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v3.5.4
making output directory... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 35 source files that are out of date
updating environment: [new config] 35 added, 0 changed, 0 removed
reading sources... [100%] triaging
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... WARNING: unsupported theme option 'copyright_url' given
done
writing output... [100%] triaging
generating indices... genindex done
writing additional pages... search done
copying images... [100%] images/python-cyclic-gc-5-new-page.png
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 1 warning.

The HTML pages are in _build/html.

Build finished. The HTML pages are in _build/html.
```

As you can see, I have got `WARNING: unsupported theme option 'copyright_url' given`

Any idea?

Thank you

CAM-Gerlach:
Thanks @matrixise . This PR neglected to add

```
copyright_url =
```

in `theme.conf` as mentioned in my review comment above, and it seems I didn't notice the omission earlier, sorry. That's why you're seeing the warning. However, it may still have worked, I'm not certain—I presume you actually checked the build output and the link was not correct?

matrixise:
> That's why you're seeing the warning. However, it may still have worked, I'm not certain—I presume you actually checked the build output and the link was not correct?

In fact, without this parameter, the build of the devguide or another project using this PR would generate a Warning with the undefined `copyright_url` option

In this case, the `copyright_url` must be defined into the `theme.conf` file.

@jablonskidev Could you update your PR with the `copyright_url` option?

Thank you

jablonskidev:
@matrixise @CAM-Gerlach Thanks for your feedback! I added `copyright_url =` to `theme.conf`. If there's anything else I can do to improve this and help move it forward, then please let me know.

matrixise:
Thanks @jablonskidev I am going to check in few minutes.

python_docs_theme/theme.conf
```diff
@@ -29,3 +29,5 @@ root_name = Python
 root_url = https://www.python.org/
 root_icon = py.png
 root_include_title = True
+
```
matrixise, line 32: Remove the empty line, it's not needed

matrixise (changes requested):

python_docs_theme/theme.conf
jablonskidev: @matrixise I removed this empty line.

matrixise (approved):

matrixise:
@Mariatta What do you think about this PR, for me, it's fine, and we can merge it.

matrixise:
@JulienPalard Just to be sure. For you, is there an issue with this PR and the "copyright" link in the existing documentations? Thanks

JulienPalard (approved):

matrixise (approved):
Thank you so much for your contribution.

PR #69, opened 2021-04-12
Test Github action to validate the theme against docsbuild scripts.

JulienPalard:
Does an admin of the repo can check if we have something to enable for actions to run?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,34 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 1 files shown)


hugovk:
I think GHA is enabled automatically, you just need to put a workflow file in `master` (e.g. by merging...).

JulienPalard:
Oh.

It make it not easy to test it before merging it on master though...

hugovk:
Indeed.

One option is to put a very basic noop workflow file in `master` first, and then a PR can test the proper one. (Or just merge and iterate from there.)

JulienPalard:
> Or just merge and iterate from there

I think I'm happy with the current state of this one.

PR #74, opened 2021-05-09
Fix the appearance of version/language selects

obulat:
This PR fixes several issues in the appearance of version and language selects in Safari and Android Chrome. 

On narrow screens, the theme overrides the native appearance of selects to make them look the same in all browsers. However, the non-prefixed `appearance: none` doesn't work in Safari and Chrome for Android, so I added the `-webkit` prefix.

I also added right padding to the version switcher to fix the overlapping of the arrow and the version number when the version number is long.

When clicking the select or focusing on it using tab, user can see an outline. Currently, it goes around the select itself, not the outer div, and looks like it's inside the box. I added the outline-offset to fix this.

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -235,6 +235,14 @@ dl > dt span ~ em {
     div.body h3, div.body h4, div.body h5 {
         font-size: 1rem;
     }
+    /* Override default styles to make more readable */
+    div.body ul {
+        padding-inline-start: 1rem;
+    }
+    div.body blockquote {
+        margin-inline-start: 1rem;
+        margin-inline-end: 0;
+    }
     /* Remove sidebar and top related bar */
     div.related, .sphinxsidebar {
         display: none;
@@ -266,6 +274,7 @@ dl > dt span ~ em {
         top: 0;
         height: 40px;
         width: 100%;
+        max-width: 100vw;
         padding: 0 1rem 0 45px;
         display: flex;
         align-items: center;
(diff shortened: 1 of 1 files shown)


obulat:
On some mobile screens, the top search form was overflowing. I fixed it by adding max-width to the navbar. This happens on pages with long method names in the body that are inside paragraphs and cannot be split, which makes the window wider than 100%.

The default paddings and margins for lists and blockquotes look good on desktop, but make the readable part of the screen too narrow on mobile, I made them narrower.

JulienPalard:
I deployed this fix on https://docs.python.org/3.11/ (english only) so it can easily be reviewed.

obulat:
> I deployed this fix on https://docs.python.org/3.11/ (english only, but other languages will automatically follow in the next 24 hours) so it can easily be reviewed.

Thank you! I will need to fix that gray background on the switchers. It isn't there when I test in devtools q

obulat:
@JulienPalard, could you deploy the latest version? I would like to see if the background color on the selects is fixed.

JulienPalard:
For the record (also using the PR comment as an extension of my own memory), here's what it take to rebuild in production:
```
docsbuild@docs:~$ /srv/docsbuild/venv-with-sphinx-3.2.1/bin/python -m pip install --upgrade git+https://github.com/obulat/python-docs-theme@patch-1
docsbuild@docs:~$ /srv/docsbuild/venv/bin/python /srv/docsbuild/scripts/build_docs.py --quick --language en --branch 3.10
```
@obulat rebuild done for 3.10 only, english only.

JulienPalard:
I'm building it agaon on docs.python.org/3.10/

[building ...]

build complete.

JulienPalard:
On docs.python.org/3.10 on my phone I'm getting a css issue: 

I don't know how to debug this in mobile but it may be the same as https://github.com/python/python-docs-theme/issues/78

JulienPalard:
Thanks \o/

PR #76, opened 2021-05-09
Remove id searchbox on mobile to fix a layout bug

obulat:
Sphinx in `doctools.js` uses an element with id `searchbox` to append a paragraph saying "Hide search matches". It is currently not used in the Python docs, but with the responsive theme, it breaks the layout. The search box at the top has a "Hide search matches" link to the right of it:

Screen Shot 2021-05-09 at 12 03 48 PM

To reproduce:
1. Go to the 3.11 version of the docs
2. Narrow your screen to mobile width
3. Search for something (eg. "xml rpc")
4. Click on one of the search results (eg. "xmlrpc.client — XML-RPC client access")

I removed the id from the search form in the top menu bar so that this link is not added.

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -90,7 +90,7 @@
          </a>
         <div class="version_switcher_placeholder"></div>
         {%- if pagename != "search" and builder != "singlehtml" %}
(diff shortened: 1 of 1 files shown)


JulienPalard:
Thanks!

PR #79, opened 2021-05-12
Improve readability

### LINKED ISSUE #77: Theme: text justification looks bad (in narrow columns)
I'm stoked to see the huge improvements of the docs on mobile. This is really fantastic.

One thing I noticed, though, is that there are often huge gaps between words. For example (on Android):
image
Using "text-align: left" would make this look a lot better.

I would go further, and recommend to never ever use "justify" at all: it generally reduces readability and often simply looks bad. Even with wider columns it often becomes a typographic disaster (Safari, desktop):
image

obulat:
Fixes #77

Replace `text-align: justify` from the classic.css with `text-align: left`
Set line-height to 1.4 instead of 130% for some elements

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -90,10 +90,10 @@ div.body {
     padding: 0 0 0 1.2em;
 }
 
-div.body p {
-    line-height: 140%;
+div.body p, div.body dd, div.body li, div.body blockquote {
+    text-align: left;
+    line-height: 1.4;
 }
-
 div.body h1, div.body h2, div.body h3, div.body h4, div.body h5, div.body h6 {
     margin: 0;
     border: 0;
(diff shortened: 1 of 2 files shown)


JulienPalard:
Beware I rebased your branch on top of master.

Oh and I tried fingerprinting the css as you told in the cache issue.

JulienPalard:
(If we keep this fingerprinting method, we'll have to describe it in the README, to try to avoid forgetting to bump this version at release time...)

JulienPalard:
Can be tested at: https://docs.python.org/3.10

JulienPalard:
Thanks @obulat \o/

ambv:
@JulienPalard, can we please release this on PyPI? I will be building 3.9.7 on Monday and would like text not to be justified.

JulienPalard:
I think this is a good idea, yes.

@Mariatta @willingc ^

(As I don't have the rights)

willingc:
@JulienPalard What's your PyPI username? I will give you rights.

PR #84, opened 2021-09-03
Exclude the floating navbar from CHM help.

### LINKED ISSUE #83: Search bar should not appear in CHM help
Since Python 3.9.6, the CHM help file has a permanent search bar at the top of the content area. When navigating to a particular location from the index or help viewer search results, the actual target location appears at the top of the content area and is covered by the search bar. This bug still exists in the CHM for 3.10rc1.

I'm not sure the presence of this search bar in the CHM help is intentional. It was added to python-docs-theme in commit 844973206 on 2021-05-07, in time for the 3.9.6 release. It cannot possibly work in CHM, and in fact, any attempt to use it leads directly to an IE error page. If it should not exist in CHM help in the first place, please disregard the rest of this bug that only describes its impact on usability.

The screenshot below was taken after double-clicking the selected index entry. You can see that the visible part of the content area starts with the first line after the method signature, but the signature itself is covered by the search bar.

Screenshot 2021-08-07 173159

The obscured strip cannot easily be made visible because there is apparently no way to scroll the content area without taking my hand off the keyboard and using the mouse to select the content area or going directly to the scroll bar.

chrullrich:
I could not get the CHM to build correctly; there was a problem with a CSS file, I think. From the generated HTML, it looks like this is correctly getting rid of the nav bar.

Fixes #83.

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -78,6 +78,7 @@
 
 {%- block body_tag %}
 {{ super() }}
+{%- if builder != 'htmlhelp' %}
 <div class="mobile-nav">
     <input type="checkbox" id="menuToggler" class="toggler__input" aria-controls="navigation"
            aria-pressed="false" aria-expanded="false" role="button" aria-label="{{ _('Menu')}}" />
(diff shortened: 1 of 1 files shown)


JulienPalard:
Thanks for the PR!

PR #90, opened 2021-11-07
Add a configuration for license URL.

### LINKED ISSUE #89: Empty history & license link in Python docs footer
From https://bugs.python.org/issue45729

This link is empty in 3.9, 3.10 and 3.11 docs:

```jinja
    {% trans pathto_license=license_file %}See <a href="{{ license_file }}">History and License</a>
    for more information.{% endtrans %}
```

I do not understand how these pathto_* things work and where they come from.  The assignemnt looks backwards to me but it is working for other links in the same footer so it’s a bit baffling.

JulienPalard:
closes #89

You're right @merwok, the assignment were reversed (and useless so they were not introducing a bug).

I made the link to the license configurable, so we could add `license_url` cpython side to point to it (near it's `issues_url` in `Doc/config.py`)

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -126,14 +126,14 @@
     <br />
     {% trans %}Examples, recipes, and other code in the documentation are additionally licensed under the Zero Clause BSD License.{% endtrans %}
     <br />
-    {% trans pathto_license=license_file %}See <a href="{{ license_file }}">History and License</a> for more information.{% endtrans %}
-    <br /><br />
+    {% if theme_license_url %}{% trans license_file=theme_license_url %}See <a href="{{ license_file }}">History and License</a> for more information.{% endtrans %}<br />{% endif %}
+    <br />
 
     {% include "footerdonate.html" %}
     <br />
 
     {% trans last_updated=last_updated|e %}Last updated on {{ last_updated }}.{% endtrans %}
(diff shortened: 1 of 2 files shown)


CAM-Gerlach:
Would make sense to have at least have a fallback to the previous behavior if the theme variables are undefined, as was implemented on #67 per my suggestion there, for backward compatibility and to follow the approach there for the copyright link? Other than that, this is a pretty important change, since the link is broken at present.

JulienPalard:
The current "fallback" is just not to display the link, I'm not sure we can do better here. Do you have something better in mind?

CAM-Gerlach:
Yeah, sorry, what I was suggesting you could do something like was done for the copyright URL right above it, fall back to a corrected version of the previous behavior, i.e. the hardcoded path, unless the existing behavior was always broken and never actually functioned as intended (which may be the case, I didn't trawl through the full commit history to check).

JulienPalard:
I'm puzzeled: the current implementation works on some pages but not all. It works at least on:

- https://docs.python.org/3.11/copyright.html
- https://docs.python.org/3.11/license.html

JulienPalard:
Hum, no, I'm just not awaken yet, on those page, it work like on many other pages: it redirects on the "Current page". In all cases it's a `<a href=""`.

JulienPalard:
I modified my PR a bit to avoid changing the msgid for translators.

PR #91, opened 2021-11-08
Make sidebar scrollable and sticky (on modern browsers)

### LINKED ISSUE #20: make sidebar sticky
All the relevant comments are over there https://bugs.python.org/issue28044 which got closed as soon as I added my comment, which isn't a particularly friendly sequence of events. If the issues there are outdated, they should be marked outdated and transferred here as part of the transition.

My comment is that the sidebar menus scroll off the top when you scroll down the page, so they are pretty useless, instead of helpful.

Nice implementation of sticky sidebar.
https://css-tricks.com/sticky-smooth-active-nav/#more-273952

I haven't looked at the patch, and don't know what R. David Murray doesn't like about how it works, but I find the Python sidebar extremely annoying because it scrolls off the top, and would rather it stayed around visible.

dlukes:
Fixes #20.

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -52,6 +52,9 @@ form.inline-search input[type="submit"] {
 }
 
 div.sphinxsidebar {
+    position: sticky;
+    top: 0;
+    height: 100vh;
     background-color: #eeeeee;
     border-radius: 5px;
     line-height: 130%;
@@ -60,6 +63,13 @@ div.sphinxsidebar {
 
 div.sphinxsidebar h3, div.sphinxsidebar h4 {
     margin-top: 1.5em;
+}
+
+div.sphinxsidebarwrapper {
+    box-sizing: border-box;
+    height: 100%;
+    overflow-x: hidden;
+    overflow-y: auto;
 }
 
 div.sphinxsidebarwrapper > h3:first-child {
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -51,10 +51,15 @@ form.inline-search input[type="submit"] {
     width: 40px;
 }
 
+div.document {
+    display: flex;
+}
+
 div.sphinxsidebar {
+    float: none;
     position: sticky;
     top: 0;
-    height: 100vh;
+    max-height: 100vh;
     background-color: #eeeeee;
     border-radius: 5px;
     line-height: 130%;
--- a/python_docs_theme/static/sidebar.js
+++ b/python_docs_theme/static/sidebar.js
@@ -2,8 +2,8 @@
  * sidebar.js
  * ~~~~~~~~~~
  *
- * This script makes the Sphinx sidebar collapsible and implements intelligent
- * scrolling. This is a slightly modified version of Sphinx's own sidebar.js.
+ * This script makes the Sphinx sidebar collapsible. This is a slightly
+ * modified version of Sphinx's own sidebar.js.
  *
  * .sphinxsidebar contains .sphinxsidebarwrapper.  This script adds in
  * .sphixsidebar, after .sphinxsidebarwrapper, the #sidebarbutton used to
@@ -26,7 +26,6 @@ $(function() {
   // the 'sidebarbutton' element is defined as global after its
   // creation, in the add_sidebar_button function
   var jwindow = $(window);
-  var jdocument = $(document);
   var bodywrapper = $('.bodywrapper');
   var documentwrapper = $('.documentwrapper');
   var sidebar = $('.sphinxsidebar');
@@ -62,8 +61,6 @@ $(function() {
       expand_sidebar();
     else
       collapse_sidebar();
-    // adjust the scrolling of the sidebar
-    scroll_sidebar();
   }
 
   function collapse_sidebar() {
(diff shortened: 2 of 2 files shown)


the-knights-who-say-ni:
Hello, and thanks for your contribution!

I'm a bot set up to make sure that the project can legally accept this contribution by verifying everyone involved has signed the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA).



## CLA Missing

Our records indicate the following people have not signed the CLA:

@dlukes

For legal reasons we need all the people listed to sign the CLA before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.

If you have recently signed the CLA, please wait at least one business day
before our records are updated.


You can [check yourself](https://check-python-cla.herokuapp.com/) to see if the CLA has been received.

Thanks again for the contribution, we look forward to reviewing it!

dlukes:
For more info, see [my comment](https://github.com/python/python-docs-theme/issues/20#issuecomment-961994763) on the associated issue.

JulienPalard:
(The buidl artifact is available here for local tests : https://github.com/python/python-docs-theme/actions/runs/1434325782)

JulienPalard:
I have a sidebar separator issue when scrolling to the bottom, see:

Screenshot 2021-11-25 at 17-54-46 3 9 9 Documentation

(Maybe this separator can just be removed now that the theme is responsive, but I think it's worth a different PR?)

dlukes:
Hmm, interesting. What OS/browser? I'm asking because I can't reproduce this on Linux on either Firefox or Chrome:

image

The sidebar still overflows a bit as a result of the switch to `box-sizing: border-box` (I think), which is definitely also worth a fix now that I see it, but something tells me it won't help with the much more dramatic overflow you seem to be experiencing.

Just to make sure: if you access <https://docs.python.org>, I suppose the sidebar renders correctly?

> (Maybe this separator can just be removed now that the theme is responsive, but I think it's worth a different PR?)

Definitely worth a separate PR, if at all :) Personally, I don't think I've ever clicked it, but I've seen comments in threads concerning the Python docs theme from people who seem very fond of it, and removing it would undoubtedly [break their workflow](https://xkcd.com/1172/).

dlukes:
OK, so right after I posted my previous reply, I accidentally stumbled upon a way to trigger the behavior you're experiencing... Sigh, typical.

It's sort of weird though -- for some reason, the sidebar gets out of whack whenever a browser UI element pops up at the bottom (???), e.g. the search bar:

image

Or the dev tools pane:

image

I've no idea how or why this is happening but I'll investigate.

(I'm guessing something funky is happening with the sidebar's JavaScript, which makes me want to get rid of it, but see above.)

dlukes:
Oh, and zooming in on the page also triggers the behavior you've described.

hugovk:
I can reproduce as well on macOS Monterey with Firefox, Chrome and Safari.

To repro I needed to reduce the height of the window:

image

Or open a page with content longer than the window height:

image

dlukes:
Sorry for the long turnaround, it's a busy time of the year, my wife and I are expecting a second daughter in a month or so, etc. :) Please take a look at the fix I just pushed, when you get a chance, and let me know if it works for you, @JulienPalard and @hugovk.

It should even fix an existing bug in the placement of the &laquo;, which is meant to stay put vertically centered in the viewport, but gets out of whack when you adjust the zoom level of the page.

Other than that, should I worry about what happens when using a browser that doesn't support viewport units? Especially w.r.t. to the collapse button -- honestly, I'd rather not put too much work into making it behave nicely on older browsers if it's going to be axed at some point in the future anyway.

JulienPalard:
(a build artefact is available for local tests in the "Checks" tab)

hugovk:
Looks good now, thanks!

image


image

dlukes:
Thanks! I've noticed another issue though -- when the content is short, the sidebar still stretches to full viewport height. I mean, that's what the CSS says, so of course it does :) Here's what it looks like, my patch on the left, current version on the right:

image

I assume we'd rather keep the current behavior? Adding so much empty space is not exactly broken, but it's definitely ugly.

dlukes:
I've pushed a fix which relies on flexbox instead of floats for the sidebar layout. Let me know what you think.

JulienPalard:
LGTM, tested on Firefox 95 (on Debian (desktop)) and on Firefox 91 (on Mobian (phone)).

I'd appreciate if someone else could take a look at it before merging, I don't trust my skills at reviewing front-end.

JulienPalard:
Thanks @dlukes for the PR !

dlukes:
Thank you for the assistance! If you have a minute, I'm just curious -- when can I expect this to land on the live website? Very roughly speaking of course -- next patch release, next minor release? Later than that?

JulienPalard:
> If you have a minute, I'm just curious -- when can I expect this to land on the live website?

I released python-docs-theme v2022.1 with your commits, so it's just the time to wait for a new build on the backend, which is govenred by [a daily cron](https://github.com/python/psf-salt/blob/6e00f136db9828c21ad05b9ba8e0928609a6996f/salt/docs/init.sls#L84), so 24h max.

In fact the cron has already picked up some, like [fr/3.9](https://docs.python.org/fr/3.9/tutorial/index.html) so you can already see it online, 24h being the time for it to be fully deployed on all versions and all languages.

dlukes:
Oh wow, great!

PR #97, opened 2022-10-16
Use pyproject

AA-Turner:
Moves to consolidated `pyproject.toml` and updates GHA.

A


.github/workflows/tests.yml
```diff
@@ -1,35 +1,29 @@
...
-          python-version: 3.8
+          python-version: 3
       - name: clone docsbuild scripts
         run: |
           git clone https://github.com/python/docsbuild-scripts/
       - name: setup requirements
         run: |
           python -m pip install --upgrade pip
-          python -m pip install jinja2 markupsafe zc.lockfile
+          python -m pip install jinja2 markupsafe zc.lockfile requests
```
hugovk, line 22: Let's use the existing requirements.txt to avoid duplication and needing to keep both in sync:

```suggestion
          python -m pip install -r docsbuild-scripts/requirements.txt
```

AA-Turner:
I copied the initial file from another project of mine that used flit, and flit gives fewer things to clean up (no egg info directory).

But the entire point of standardisation is that changing the build backend should be easy so if you'd prefer setuptools we could change?

A

JulienPalard:
> changing the build backend should be easy

It is and that's good.

What I fear is build backends created "just because it's possible", maintained by a single person and backends with tons of lines of code. I don't know them all but for example pipenv was 304k lines of code last time I checked. As I like to say E=MC² (Errors = MoreCode²).

I just checked at flit, it's 4k lines of code, 2k lines of tests, which sounds good to me. It has a single maintainer though, which does not sounds good (but setuptools don't have many more...).

Facing this "big mess of new backends", and as setuptools is handling pyproject.toml correctly, my reaction is to try to stick to the standard. (I see setuptools as the "de facto standard" and other backends as "side projects", I know this is a very personal and simplified point of view).

Flit looks simple and well maintained, LGTM, don't bother changing it.

ezio-melotti (approved):

hugovk (approved):

hugovk:
Thanks!

PR #94, opened 2022-05-30
Update GitHub Actions testing workflow

AA-Turner:
Updated remaining jQuery scripts to standard JavaScript, and a few fixes for CSS changes.

I marked various backwards compatability points explicitly -- I don't know what the supported Sphinx versions are for this theme.

A

PR #99, opened 2022-11-11
Update ``python-docs-theme`` to work with Sphinx 5 & 6

AA-Turner:
Updated remaining jQuery scripts to standard JavaScript, and a few fixes for CSS changes.

A

### DIFF
--- a/python_docs_theme/static/copybutton.js
+++ b/python_docs_theme/static/copybutton.js
@@ -1,64 +1,70 @@
-$(document).ready(function() {
+
+const loadCopyButton = () => {
     /* Add a [>>>] button on the top-right corner of code samples to hide
      * the >>> and ... prompts and the output and thus make the code
      * copyable. */
-    var div = $('.highlight-python .highlight,' +
-                '.highlight-python3 .highlight,' +
-                '.highlight-pycon .highlight,' +
-                '.highlight-pycon3 .highlight,' +
-                '.highlight-default .highlight');
-    var pre = div.find('pre');
+    const hide_text = "Hide the prompts and output"
+    const show_text = "Show the prompts and output"
 
-    // get the styles from the current theme
-    pre.parent().parent().css('position', 'relative');
-    var hide_text = 'Hide the prompts and output';
-    var show_text = 'Show the prompts and output';
-    var border_width = pre.css('border-top-width');
-    var border_style = pre.css('border-top-style');
-    var border_color = pre.css('border-top-color');
-    var button_styles = {
-        'cursor':'pointer', 'position': 'absolute', 'top': '0', 'right': '0',
-        'border-color': border_color, 'border-style': border_style,
-        'border-width': border_width, 'color': border_color, 'text-size': '75%',
-        'font-family': 'monospace', 'padding-left': '0.2em', 'padding-right': '0.2em',
-        'border-radius': '0 3px 0 0'
-    }
-
-    // create and add the button to all the code blocks that contain >>>
-    div.each(function(index) {
-        var jthis = $(this);
-        if (jthis.find('.gp').length > 0) {
-            var button = $('<span class="copybutton">&gt;&gt;&gt;</span>');
-            button.css(button_styles)
-            button.attr('title', hide_text);
-            button.data('hidden', 'false');
-            jthis.prepend(button);
-        }
-        // tracebacks (.gt) contain bare text elements that need to be
-        // wrapped in a span to work with .nextUntil() (see later)
-        jthis.find('pre:has(.gt)').contents().filter(function() {
-            return ((this.nodeType == 3) && (this.data.trim().length > 0));
-        }).wrap('<span>');
-    });
-
-    // define the behavior of the button when it's clicked
-    $('.copybutton').click(function(e){
-        e.preventDefault();
-        var button = $(this);
-        if (button.data('hidden') === 'false') {
+    const button = document.createElement("span")
+    button.classList.add("copybutton")
+    button.innerText = ">>>"
+    button.title = hide_text
+    button.dataset.hidden = "false"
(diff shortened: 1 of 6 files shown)


AA-Turner:
Turns out this is required for Sphinx 5... https://github.com/python/cpython/pull/99381

A

pradyunsg (approved):
LGTM

python_docs_theme/static/copybutton.js
```diff
@@ -1,64 +1,63 @@
...
+            codeEl.querySelectorAll('.go, .gp, .gt').forEach(el => el.hidden = true)
+            codeEl.querySelectorAll('.gt').forEach(el => {
+                while ((el = el.nextSibling) && el.nodeType !== Node.DOCUMENT_NODE) {
+                    if (el.nodeType === Node.ELEMENT_NODE && el.matches(".gp, .go")) break;
+                    if (el.nodeType === Node.TEXT_NODE && el.textContent.trim()) {
+                        const wrapper = document.createElement('span');
+                        el.after(wrapper);
+                        wrapper.appendChild(el);
+                        el = wrapper
+                    }
```
ezio-melotti, line 30: Why is this necessary?

python_docs_theme/static/copybutton.js
```diff
@@ -1,64 +1,63 @@
...
+            buttonEl.title = show_text
+            buttonEl.dataset.hidden = "true"
         } else {
             // show the code output
-            button.parent().find('.go, .gp, .gt').show();
-            button.next('pre').find('.gt').nextUntil('.gp, .go').css('visibility', 'visible');
-            button.css('text-decoration', 'none');
-            button.attr('title', hide_text);
-            button.data('hidden', 'false');
+            codeEl.childNodes.forEach(el => el.hidden = false)
```
ezio-melotti, line 38: IIUC here you are just assuming that everything inside the code block was already visible by default, and you are setting `.hidden = false` to each element.  Do you think this might conflict with other things (e.g. Sphinx extensions) that might introduce hidden elements?

ezio-melotti (commented):
In addition to the inline comments, I have two generic comments:
* The use of semicolons at the end of the line seems inconsistent.  Unless there is a specific reason to keep them, I would just remove them all.
* I would avoid the use of `if (...) return;` or `if (...) break;` or even `if (...) some_functions()`.  The expression after the condition should go on a new line, and it should be surrounded by `{ }`, e.g.:
   ```js
   if (...) {
      ...
   }
   ```

---

Fun fact: I originally implemented both the collapsible sidebar (~12 years ago in https://github.com/python/cpython/issues/47393) and the copybutton (~11 years ago in https://github.com/python/cpython/commit/59dba38166160163ce4ac473933ab1c74b6222de) and since then they have traveled around a number of dirs and repos.  It's nice to see that in the meanwhile things got standardized and JS evolved to the point that JQuery is no longer needed.

.github/workflows/tests.yml
```diff
@@ -1,35 +1,39 @@
----
-
 name: Tests
```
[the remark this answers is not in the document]
AA-Turner: This branch is based on top of #94, which is required for CI to pass as of the change to required dependencies -- hopefully #94 is merged soon and I can rebase.

A

python_docs_theme/static/copybutton.js
AA-Turner: Yes, this replicates the current behaviour -- I would suggest merging this as-is to keep the behaviour consistent and if the unconditional hiding is a problem we can change it later.

A

python_docs_theme/static/copybutton.js
AA-Turner: I've added a note -- in efect tracebacks have bare text elements that we need to be able to hide, so we wrap in a span. This replicates the current behaviour.

A

python_docs_theme/static/copybutton.js
ezio-melotti: The original implementation was using `button.parent().find('.go, .gp, .gt')` `button.next('pre').find('.gt').nextUntil('.gp, .go')` to select the elements, but in this version it seems you are doing this only in the `if` branch but not in the `else`.

Also, without JQuery, selecting the tracebacks seems more convoluted, so I understand you might not want to duplicate it in the `else`, but perhaps you could create a function to show/hide tracebacks and call it in both the `if` and the `else`?  This will make the implementation closer to the original and also more readable/understandable.

AA-Turner: I've added a new `getHideableCopyButtonElements` generator function.

A

hugovk:
Is this ready to merge?

hugovk:
Rebased on `main` to build RTD preview: 

* https://python-docs-theme-previews--99.org.readthedocs.build/en/99/

And there's an error in the console:

```
sidebar.js:34 Uncaught TypeError: Cannot read properties of null (reading 'querySelector')
    at HTMLDocument.initialiseSidebar (sidebar.js:34:38)
```

From the last line of:

```js
  const bodyWrapper = document.getElementsByClassName("bodywrapper")[0]
  const sidebar = document.getElementsByClassName("sphinxsidebar")[0]
  const sidebarWrapper = document.getElementsByClassName('sphinxsidebarwrapper')[0]
  const sidebarButton = document.getElementById("sidebarbutton")
  const sidebarArrow = sidebarButton.querySelector('span')
```

AA-Turner:
Updated the JavaScript, the previous version was attempting to access an element that had yet to be created. Thank you for the review!

A

hugovk (approved):
Working fine now, thanks!

JulienPalard:
I just built https://docs.python.org/3.12/ after merging this PR.

JulienPalard:
https://github.com/python/python-docs-theme/pull/126/files

Issue #78, opened 2021-05-10
Theme issue (or Firefox cache issue?)


JulienPalard:
Today I'm seeing a theming issue:

![Screenshot_2021-05-10 Built-in Functions — Python 3 10 0b1 documentation](https://user-images.githubusercontent.com/239510/117622984-86737100-b173-11eb-9582-f5f1e2c4aa50.png)

It looks like cache problem, so I'm going to investigate step by step...

...damned, opening the Firefox console solved the issue (the no-cache were ticked), so we have a browser cache issue: users with the old theme in their local cache may get what I got.

I unticked the Firefox network console "disable cache" and tried all other translations to try to reproduce, and was able to reproduce it once:

The request done by my browser was:

```
curl 'https://docs.python.org/zh-cn/3.10/_static/pydoctheme.css' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:88.0) Gecko/20100101 Firefox/88.0' -H 'Accept: text/css,*/*;q=0.1' -H 'Accept-Language: fr,en-US;q=0.7,en;q=0.3' --compressed -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Referer: https://docs.python.org/zh-cn/3.10/library/functions.html' -H 'Sec-GPC: 1' -H 'TE: Trailers'
```

(Which, using curl, give the up-to-date file)

But got no response (first time I see this) it were "raced" (by the local cache?) I only see `882 B (raced)` in the network pane, no more info.

According to https://support.mozilla.org/en-US/questions/1267945 this display mean the network won the race, which looks wrong because I have no response header and the css is the old one:

![cacheissue](https://user-images.githubusercontent.com/239510/117624333-17971780-b175-11eb-9640-4d4a3e78eac2.png)

Also the size is too low to contain the whole css, so it looks like the network connection has started, bytes has been received, but the cache won the race and the network connection was interrupted.

Which I find strange is that when I refresh, all other components are requested by my browser, which all properly receive a "304 not modified" answer, but not the CSS itself:

![cacheissue2](https://user-images.githubusercontent.com/239510/117625242-1b776980-b176-11eb-818b-f5784fe9fa63.png)

JulienPalard:
Now "pydoctheme.css" completly dissapeared on my network pane.... opened a Firefox issue https://bugzilla.mozilla.org/show_bug.cgi?id=1710373 about this.

JulienPalard:
[...]

PR #108, opened 2023-02-16
Append a hash ?digest to CSS files for cache-busting

hugovk:
Fix https://github.com/python/python-docs-theme/issues/78.

Append a `?digest=hash` to the end of the `pydoctheme.css`, computed from the file contents, so when a new CSS file is deployed, the old one is no longer used from the browser cache.

For example:

`<link rel="stylesheet" type="text/css" href="_static/pydoctheme.css?digest=afc8307635b40ad4bb21df93e5fc348bcdad7f27" />`

This is based on how @pradyunsg's Furo theme does it:

https://github.com/pradyunsg/furo/blob/193643fdb6787501195555244f4a9e953ef544bb/src/furo/__init__.py#L149-L161

# Demo

View the source of pages at https://python-docs-theme-previews--108.org.readthedocs.build/en/108/

### DIFF
--- a/python_docs_theme/__init__.py
+++ b/python_docs_theme/__init__.py
@@ -1,10 +1,61 @@
+import hashlib
 import os
+from functools import lru_cache
+from pathlib import Path
+from typing import Any, Dict, List
+
+import sphinx.application
+from sphinx.builders.html import StandaloneHTMLBuilder
+
+THEME_PATH = Path(__file__).parent.resolve()
+
+
+@lru_cache(maxsize=None)
+def _asset_hash(path: str) -> str:
+    """Append a `?digest=` to an url based on the file content."""
+    full_path = THEME_PATH / path.replace("_static/", "static/")
+    digest = hashlib.sha1(full_path.read_bytes()).hexdigest()
+
+    return f"{path}?digest={digest}"
+
+
+def _add_asset_hashes(static: List[str], add_digest_to: List[str]) -> None:
+    for asset in add_digest_to:
+        index = static.index(asset)
+        static[index].filename = _asset_hash(asset)  # type: ignore
+
+
+def _html_page_context(
+    app: sphinx.application.Sphinx,
+    pagename: str,
+    templatename: str,
+    context: Dict[str, Any],
+    doctree: Any,
+) -> None:
+    if app.config.html_theme != "python_docs_theme":
+        return
+
+    assert isinstance(app.builder, StandaloneHTMLBuilder)
+
+    if "css_files" in context:
+        if "_static/pydoctheme.css" not in context["css_files"]:
+            raise ValueError(
+                "This documentation is not using `pydoctheme.css` as the stylesheet. "
+                "If you have set `html_style` in your conf.py file, remove it."
+            )
+
+        _add_asset_hashes(
+            context["css_files"],
+            ["_static/pydoctheme.css"],
+        )
 
 
 def setup(app):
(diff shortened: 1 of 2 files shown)


pradyunsg (approved):

m-aciek:
This changes break backwards compatibility with Sphinx 3.4.3 which is pinned for Python 3.10 documentation.
Following is seen in the stdout when trying to build:
```
% make html
…
Extension error:
Handler <function _html_page_context at 0x110466020> for event 'html-page-context' threw an exception (exception: This documentation is not using `pydoctheme.css` as the stylesheet. If you have set `html_style` in your conf.py file, remove it.)
```
Does the same for compatibility with Sphinx 2.4.4 pinned for Python 3.9 and Python 3.8 docs:
```
% make html
…
                                                                                                                                                                         
Exception occurred:
  File "/Users/maciej.olko/projects/cpython/Doc/venv/lib/python3.12/site-packages/python_docs_theme/__init__.py", line 42, in _html_page_context
    raise ValueError(
    ^^^^^^^^^^^^^^^^^
ValueError: This documentation is not using `pydoctheme.css` as the stylesheet. If you have set `html_style` in your conf.py file, remove it.
The full traceback has been saved in /var/folders/c0/vxg8dn6n28105qpdxmg3kblc0000gq/T/sphinx-err-psaoa2co.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

(Theme seems compatible with Sphinx 2.3.1 which is used for Python 3.7.)

Would it be possible to tweak the theme to support Sphinx 3.4.3 and 2.4.4 for older Python docs?

hugovk:
Thanks for reporting.

Sphinx 3.4.3: `context["css_files"]=[]`

Sphinx 6.1.3: `context["css_files"]=['_static/pygments.css', '_static/pydoctheme.css']`

---

Options include pinning `python-docs-theme`, or ideally upgrading Sphinx for the other version branches.

But from a practical maintenance point of view, shall we just skip this for older Sphinx versions?


```diff
commit 8b88430553fdaa1d7a85296a106158b5faf99a4b
Author: Hugo van Kemenade <hugovk@users.noreply.github.com>
Date:   Sat Mar 11 13:32:42 2023 +0200

    Skip cache-busting for old Sphinx

diff --git a/python_docs_theme/__init__.py b/python_docs_theme/__init__.py
index 62b16f0..bbe1352 100644
--- a/python_docs_theme/__init__.py
+++ b/python_docs_theme/__init__.py
@@ -37,7 +37,7 @@ def _html_page_context(
 
     assert isinstance(app.builder, StandaloneHTMLBuilder)
 
-    if "css_files" in context:
+    if sphinx.version_info >= (4,) and "css_files" in context:
         if "_static/pydoctheme.css" not in context["css_files"]:
             raise ValueError(
                 "This documentation is not using `pydoctheme.css` as the stylesheet. "
```

m-aciek:
I think skipping is most pragmatic solution. 👍  I was also thinking about pinning, but the downside is it would prevent users from having another fixes from this release. When having similar conversation earlier, upgrading Sphinx wasn't told to be good option for already published stable releases, because of Linux distributions that provide Sphinx as a system package, if I recall correctly.

hugovk:
Please see PR https://github.com/python/python-docs-theme/pull/113.

PR #113, opened 2023-03-11
Skip cache-busting for old Sphinx

hugovk:
Fix for https://github.com/python/python-docs-theme/pull/108#issuecomment-1464879086.

Skip cache busting for Sphinx 3 and older.


m-aciek (approved):
I've tested locally with Python 3.10 (Sphinx 3.4.3) and Python 3.7 (Sphinx 2.3.1), building and looking fine. Thank you!

PR #114, opened 2023-03-11
Add matrix strategy to test with all supported Python versions

m-aciek:


pradyunsg (approved):

pradyunsg:
I'll go ahead and merge this, on the basis that this seems like a good idea overall. Holler if you spot something off! :)

hugovk:
Is it possible to add 3.12/`main`?

Should the `doc-html` artifact at https://github.com/python/python-docs-theme/actions/runs/4393087253 also include 3.8-3.10?

Left is `main`, right is this PR:

image

m-aciek:
> Is it possible to add 3.12/main?

Yes, definitely, I think it's a good idea, I can add a pull request for that.

> Should the doc-html artifact at https://github.com/python/python-docs-theme/actions/runs/4393087253 also include 3.8-3.10?

Actually, we should see the jobs for 3.8-3.10 as failed, as my branch wasn't basing on fresh-fixed master. The missing artifacts were because of failed jobs. It looks like exit code of `build_docs.py` script is incorrectly zero instead of non-zero for failing build. I will try to confirm that and follow it up in builddocs-script GitHub repository.

That's what the artifacts from the Actions on the theme main branch look like for me (all versions from matrix are there):

Zrzut ekranu 2023-03-11 o 21 16 18

m-aciek:
Please see PR https://github.com/python/docsbuild-scripts/pull/150.

m-aciek:
I just wanted to share a thought that in case of a need of some changes that would break backwards compatibility we can always decide to pin the theme version in CPython branch and remove a version from matrix. But still it will be easier to spot the breakage and pin the version in advance of releasing such changes having those tests in CI.

PR #115, opened 2023-03-12
Test also with main branch of Python

m-aciek:
As suggested by @hugovk https://github.com/python/python-docs-theme/pull/114#issuecomment-1464960658. Follow-up for #114.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,11 +7,11 @@ on:
 
 jobs:
   build_doc:
-    name: Build CPython documentation
+    name: Build CPython docs
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: [3.7, 3.8, 3.9, '3.10', 3.11]
+        branch: [origin/main, 3.11, '3.10', 3.9, 3.8, 3.7]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
@@ -35,7 +35,7 @@ jobs:
           --skip-cache-invalidation 
           --theme $(pwd) 
           --language en 
-          --branch ${{ matrix.version }}
+          --branch ${{ matrix.branch }}
       - name: Upload
         uses: actions/upload-artifact@v3
         with:


ezio-melotti (approved):

.github/workflows/tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: [3.7, 3.8, 3.9, '3.10', 3.11]
+        branch: [3.7, 3.8, 3.9, '3.10', 3.11, origin/main]
```
hugovk, line 14: I think this can also be a version number?

```suggestion
        branch: [3.7, 3.8, 3.9, '3.10', 3.11, 3.12]
```

A quick test adding `sys.exit(versions)` after `versions = Version.filter(VERSIONS, args.branch)`:

https://github.com/python/docsbuild-scripts/blob/d5a0a72b373c81f64bbabcbe8b885ba2a409be38/build_docs.py#L1052

gives the same results:

```console
$ python build_docs.py --branch origin/main
[Version(3.12)]
$ python build_docs.py --branch 3.12
[Version(3.12)]
```

ezio-melotti: One advantage of using `origin/main` is that it shows more explicitly that we want to test on the currently-in-development branch, and not just the latest (released) version of Python.  If we use `3.12`, once 3.12 gets released people might forget to add `3.13` and think that the list is already up to date.
If we decide to use `3.12`, a comment should be added to explicitly mention that.

I'm also not sure how `3.12` currently work, since there's no branch/tag called `3.12` afaict.  Does it automatically fall back on `main` if the branch is not found?  Does this raise any warning?

hugovk: > One advantage of using `origin/main` is that it shows more explicitly that we want to test on the currently-in-development branch, and not just the latest (released) version of Python. If we use `3.12`, once 3.12 gets released people might forget to add `3.13` and think that the list is already up to date. If we decide to use `3.12`, a comment should be added to explicitly mention that.

Sure.

> I'm also not sure how `3.12` currently work, since there's no branch/tag called `3.12` afaict. Does it automatically fall back on `main` if the branch is not found? Does this raise any warning?

It goes through this `filter` function:

```python
    @staticmethod
    def filter(versions, branch=None):
        """Filter the given versions.
        If *branch* is given, only *versions* matching *branch* are returned.
        Else all live version are returned (this mean no EOL and no
        security-fixes branches).
        """
        if branch:
            return [v for v in versions if branch in (v.name, v.branch_or_tag)]
        return [v for v in versions if v.status not in ("EOL", "security-fixes")]
```

https://github.com/python/docsbuild-scripts/blob/d5a0a72b373c81f64bbabcbe8b885ba2a409be38/build_docs.py#L141-L151

Where `versions` is `VERSIONS`:

```python
# EOL and security-fixes are not automatically built, no need to remove them
# from the list, this way we can still rebuild them manually as needed.
#
# Please keep the list in reverse-order for ease of editing.
VERSIONS = [
    Version("3.12", branch="origin/main", status="in development"),
    Version("3.11", branch="origin/3.11", status="stable"),
    Version("3.10", branch="origin/3.10", status="stable"),
    Version("3.9", branch="origin/3.9", status="security-fixes"),
    Version("3.8", branch="origin/3.8", status="security-fixes"),
    Version("3.7", branch="origin/3.7", status="security-fixes"),
    Version("3.6", tag="3.6", status="EOL"),
    Version("3.5", tag="3.5", status="EOL"),
    Version("2.7", tag="2.7", status="EOL"),
]
```

https://github.com/python/docsbuild-scripts/blob/d5a0a72b373c81f64bbabcbe8b885ba2a409be38/build_docs.py#LL206-L220

So it can match on either the `"3.12"` name or `"origin/main"` branch.

hugovk (approved):

m-aciek:
I've reordered branches in the matrix and shortened the job name for better visiblity/layout.

Zrzut ekranu 2023-03-14 o 21 59 47

hugovk:
Thanks!

PR #121, opened 2023-03-25
Override font for .sig for consistency with other code blocks

Kwpolska:
The `basic.css` Sphinx stylesheet introduces a `.sig` rule, which sets some monospace fonts for the signature. The `pydoctheme.css` stylesheet has its own list of fonts, applied to some other elements, but not `.sig`. The result is an inconsistency on Windows systems with both Cascadia Mono and Consolas installed (I believe that happens in a default Windows 11 install). Compare the glyphs for `64` in the function signature and in the code block below it:

image

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -255,7 +255,8 @@ div.footer a:hover {
     color: #229;
 }
 
-dl > dt span ~ em {
+dl > dt span ~ em,
+.sig {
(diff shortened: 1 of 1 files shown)


Mariatta:
Would you mind showing the "before" and "after" screenshots of how it looks like with/without this change? Thank you.

Kwpolska:
Sure. [Before](https://python-docs-theme-previews--123.org.readthedocs.build/en/123/library/base64.html#base64.encode) on the left, [after](https://python-docs-theme-previews--121.org.readthedocs.build/en/121/library/base64.html#base64.encode) on the right:

image

hugovk (approved):

PR #124, opened 2023-04-12
Remove searchbox ID from form

### LINKED ISSUE #117: Dark theme: "Hide Search Matches" shown on mobile
1. Using the new dark theme (#44) in mobile layout: https://docs.python.org/dev/
2. Search for something (e.g. "turtle")

image

3. Click a result

image

Actual result: "Hide Search Matches" is shown subscript to the right of the search box

image

Expected result: It's either shown more neatly, or, like desktop view, isn't shown at all:

image

cc @septatrix

septatrix:
Fixes #117.

The "Hide Search Matches" was not properly integrated into the layout, removing the ID reverts it to the previous behaviour where the toggle was suppressed.

Ideally we would have this working properly, however, sphinx sets this based on ID and we have more than one search forms so this would not show up properly (at the top and bottom of the desktop layout, and one in the mobile layout). With some CSS trickery it might be possible to reuse a single form element for the desktop and mobile layout such that this could be reasonably supported (and do without the toggle for the search form at the bottom) though this is a decent bit more complex and I am currently short on time.

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -103,7 +103,7 @@
             </a>
             <span class="version_switcher_placeholder"></span>
             {%- if pagename != "search" and builder != "singlehtml" %}
-            <form id="searchbox" role="search" class="search" action="{{ pathto('search') }}" method="get">
+            <form role="search" class="search" action="{{ pathto('search') }}" method="get">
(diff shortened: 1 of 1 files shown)


hugovk (approved):
Thanks, looks good:

image

PR #127, opened 2023-04-20
Don't duplicate the sidebar button on Sphinx 5 and newer

AA-Turner:
The sidebar button is included in the generated HTML from Sphinx 5 onwards (specifically, from https://github.com/sphinx-doc/sphinx/commit/3b01fbe2adf3077cad5c2cf345c6d000d429d7ac and the changes to ``themes/classic/layout.html``).

We use  the ``sphinx_version_tuple`` templating variable addded in Sphinx 4.2 to determine if we should add the sidebar button dynamically via JavaScript, or load the pre-existing button element from HTML.

I have tested locally on Sphinx 6.1.3, Sphinx 3.2.0, and Sphinx 4.5.0--all three display correct behaviour with respect to the sidebar.

A


python_docs_theme/static/sidebar.js_t
```diff
@@ -36,6 +36,16 @@ const initialiseSidebar = () => {
...
+{# Check if we need to dynamically insert the sidebar button.
+ # We prefer the ``sphinx_version_tuple`` variable, and if it is undefined we
+ # know we are running a Sphinx version older than 4.2.
+ #
+ # See: https://www.sphinx-doc.org/en/master/development/templating.html#sphinx_version_tuple
+ #}
+{% if sphinx_version_tuple is defined and sphinx_version_tuple[0] >= 5 %}
+  const sidebarButton = document.getElementById("sidebarbutton")
+  const sidebarArrow = sidebarButton.querySelector('span')
+{%  else %}
```
AA-Turner, line 48:
```suggestion
{% else %}
```

AA-Turner:
I've updated https://github.com/python/cpython/pull/99381 to point to this PR, so that it installs Sphinx 6.1.3 and `python-docs-theme` as of this PR. The [deploy preview](https://deploy-preview-99381--python-cpython-preview.netlify.app/) seems to show that all is working as intended on newer Sphinx versions, and this PR's RTD preview shows it working on older Sphinx versions.

A

pradyunsg (approved):

hugovk (approved):

AA-Turner:
Thanks Hugo!

A

PR #133, opened 2023-07-07
Dark mode: fix contrast for C++ specific styling

hugovk:
# Before

1. Visit https://docs.python.org/3.12/c-api/arg.html#numbers
2. Look at the C keywords such as `unsigned char`

<table>
<tr>
 <td>
 <td>
</table>

3. Dark mode: the dark blue text on dark background is very hard to read
4. Check the contrast ratio in devtools, it's below the WCAG AA guidelines of 4.5:

image

5. (Light mode is fine, ratio is 9.88, above the AAA ratio of 7.0.)

# After

That is defined by Sphinx: https://github.com/sphinx-doc/sphinx/blob/d3c91f951255c6729a53e38c895ddc0af036b5b9/sphinx/themes/basic/static/basic.css_t#L547-L550

Copy that CSS to our dark mode file and use `#5283ff` as suggested by devtools, and the dark mode is improved, with no change to light mode:

<table>
<tr>
 <td>

 <td>

</table>

# Preview

https://python-docs-theme-previews--133.org.readthedocs.build/en/133/c-api/arg.html#numbers

### DIFF
--- a/python_docs_theme/static/pydoctheme_dark.css
+++ b/python_docs_theme/static/pydoctheme_dark.css
@@ -124,3 +124,12 @@ div.note {
 img.invert-in-dark-mode {
     filter: invert(1) hue-rotate(.5turn);
 }
+
+/* -- object description styles --------------------------------------------- */
+
+/* C++ specific styling */
+
+.sig.c   .k, .sig.c   .kt,
+.sig.cpp .k, .sig.cpp .kt {
+    color: #5283ff;
+}
--- a/python_docs_theme/static/pydoctheme_dark.css
+++ b/python_docs_theme/static/pydoctheme_dark.css
@@ -129,6 +129,7 @@ img.invert-in-dark-mode {
 
 /* C++ specific styling */
 
+/* Override Sphinx's basic.css to fix colour contrast */
 .sig.c   .k, .sig.c   .kt,
 .sig.cpp .k, .sig.cpp .kt {
     color: #5283ff;


hugovk:
This will also fix some of the Lighthouse Accessibility audits findings.

# Before

image

# After

image

python_docs_theme/static/pydoctheme_dark.css
```diff
@@ -124,3 +124,12 @@ div.note {
...
 }
+
+/* -- object description styles --------------------------------------------- */
+
+/* C++ specific styling */
+
+.sig.c   .k, .sig.c   .kt,
+.sig.cpp .k, .sig.cpp .kt {
+    color: #5283ff;
+}
```
ezio-melotti, line 135: Since the goal is fixing contrast, I would mention it in a comment so that whoever edits the theme colors next is aware of it.

hugovk:
Thanks for bringing it to my attention, I'd not heard of APCA before.

There's an interesting discussion about it at https://github.com/jupyter/accessibility/issues/65 which also shares this article: https://kilianvalkhof.com/2021/accessibility/wcag-2-is-what-we-have/

Those conclusions seems reasonable: WCAG 2 is a standard agreed through consensus. WCAG 3 is still a draft, and is likely many years from consensus and approval. The consensus is important, it shows there's wide support, consistency, and can be broadly applied. WCAG 2 AA should be used as a baseline. 

For example, the EU's Web Accessibility Directive references WCAG 2.1 at level AA.

(Lots more detail: https://digital-strategy.ec.europa.eu/en/policies/web-accessibility-directive-standards-and-harmonisation)

(And WCAG 2 is consistently identified and upheld for the Americans with Disabilities Act https://www.boia.org/blog/is-there-a-legal-requirement-to-implement-wcag)

Level AAA includes AA, and [W3C say of AAA](https://www.w3.org/TR/WCAG21/#cc1):

> It is not recommended that Level AAA conformance be required as a general policy for entire sites because it is not possible to satisfy all Level AAA Success Criteria for some content.

AAA is a very high bar and often seen as requiring too much of a change to a design, compared with AA. I usually check the AAA colours too, but often they're a bit too much washed out, so stick with AA.

So I think using WCAG 2.1 AA should be seen as a baseline, and we can of course do better than this where appropriate.

As it happens, the suggested `#8ab7ff` is also what Chrome suggests for AAA:

image

Here's a comparison of the three options:

<table>
<tr>
 <td>Original: 1.6
 <td>AA: 4.59
 <td>AAA: 7.09
<tr>
 <td>
image

 <td>
image

 <td>
image

</table>

There's definitely a big improvement from the original -> AA.

The AAA seems a bit on the light side compared to AA, but I guess it's not too bad?

It's fairly similar to the `#77aaff` used for `Py_ssize_t`, `bytes` and `bytearray` at the bottom of the screen (which has ratio of 6.8, just below AAA).

What do you think?

hugovk:
Thanks for the discussion! Let's go with this, and further improvements welcome.

Myndex:
Hi @hugovk and @sorcio 

Under WCAG 2.x, If you have one color theme that is compliant, any number of other color schemes may use alternate rules, provided there is a means for the user to switch them.

This is important for dark mode, as WCAG 2 can not calculate for dark mode.

APCA can calculate for dark mode—regarding sorcio's comment:

> ..._lower contrast can be desirable by users of dark themes_...

Yes, in dark environments dark mode with an overall dark screen and in a dark room or at night, can use a different contrast. At the moment this is not in the draft guidelines, which are based on the more difficult case of a brightly it ambient.

There are some unreleased versions of APCA with multiple color inputs for this purpose.

APCA exceeds WCAG 2.x in actual accessibility, though there are some edge cases where there is no backwards compatibility, due to the fact that WCAG 2.x is substantially incorrect in calculating dark color pairs.

ALSO, there is a drop in replacement for WCAG 2 that is fully backwards compatible, but using APCA technology, called BridgePCA, see [bridgepca.com](https://bridgepca.com)

PR #131, opened 2023-06-15
Focus search box when pressing slash

hugovk:
Fixes https://github.com/python/python-docs-theme/issues/130.

Many websites focus the search box when the forward slash <kbd>/</kbd> key is pressed.

For example:

* Furo theme https://pradyunsg.me/furo/quickstart/
* PyPI https://pypi.org/project/python-docs-theme/
* GitHub https://github.com/python/python-docs-theme/issues/130
* YouTube https://www.youtube.com/watch?v=nMekFX2CDVk
* Discord https://discuss.python.org/

Let's do the same.

# Demo

https://python-docs-theme-previews--131.org.readthedocs.build/en/131/

PR #135, opened 2023-07-25
Enable search shortcuts

miketheman:
Enable already-existing JavaScript flag.

Included since Sphinx 4.5.0 https://www.sphinx-doc.org/en/master/changes.html#release-4-5-0-released-mar-28-2022

The [doctools.js file that is served today](https://docs.python.org/3/_static/doctools.js) includes the functionality - look for `DOCUMENTATION_OPTIONS.ENABLE_SEARCH_SHORTCUTS`

I _think_ that the only piece needed is setting the config option [`enable_search_shortcuts = True`](https://github.com/sphinx-doc/sphinx/blob/5cf3dce36ec35c429724bf1312ece9faa0c8db39/sphinx/themes/basic/theme.conf#L13C1-L14C1) in https://github.com/python/python-docs-theme/blob/4dbc102a556c520b10fc387821967e54dee77654/python_docs_theme/theme.conf

We inherit from [`default` theme](https://github.com/sphinx-doc/sphinx/blob/5cf3dce36ec35c429724bf1312ece9faa0c8db39/sphinx/themes/default/theme.conf#L2), which [inherits `classic`](https://github.com/sphinx-doc/sphinx/blob/5cf3dce36ec35c429724bf1312ece9faa0c8db39/sphinx/themes/classic/theme.conf#L2), which inherits [`basic` theme](https://github.com/sphinx-doc/sphinx/blob/5cf3dce36ec35c429724bf1312ece9faa0c8db39/sphinx/themes/basic/theme.conf#L1-L2).

Alternative to #131

### DIFF
--- a/python_docs_theme/theme.conf
+++ b/python_docs_theme/theme.conf
@@ -33,3 +33,4 @@ root_url = https://www.python.org/
 root_icon = py.svg
 root_include_title = True
 copyright_url =
+enable_search_shortcuts = True


miketheman:
Sadly, this doesn't work 100% because we create multiple searchbox elements per page with our own layout.

https://github.com/python/python-docs-theme/blob/4dbc102a556c520b10fc387821967e54dee77654/python_docs_theme/layout.html#L51
https://github.com/python/python-docs-theme/blob/4dbc102a556c520b10fc387821967e54dee77654/python_docs_theme/layout.html#L110
(there's one more, I haven't hunted it down yet)

I'm guessing this would be a little more involved in restructuring the layout template to reduce the duplication, and thus enable taking advantage of their already-supplied functionality. If there's appetite for that, I'll pursue it.

hugovk:
> Alternative to #131

Nice, I didn't know this was a Sphinx option, I'd prefer to use this than the custom JavaScript in #131. (Although there's not much JS in there.)

https://www.sphinx-doc.org/en/master/usage/theming.html says:

> **enable_search_shortcuts** (true or false): Allow jumping to the search box with `/` and allow removal of search highlighting with `Esc`.
> 
> Defaults to `True`.

Does it really default to `True`?

We don't have `/` jumping, but we do have `Esc` clearing. Where is the `Esc` coming from?

miketheman:
Turns out, it does default to True, so this PR is also unnecessary!
That's why the Esc key works already. 

Pressing the slash key today does in fact work, however it places the focus on the hidden mobile nav search box since it appears first in the HTML.

🤔

I wonder if we still need an explicit mobile nav, or if the signature could be changed so it still works correctly but does not appear first in the selector.

miketheman:
Even simpler! https://github.com/python/python-docs-theme/pull/141

Issue #129, opened 2023-05-29
Wrap lines of code to prevent empty horizontal scroll


hugovk:
# Steps to reproduce

1. On mobile, go to https://docs.python.org/3.13/whatsnew/3.12.html

I'm using Android / Samsung S20 / Chrome, can also repro in macOS / Chrome desktop emulating mobile with at least iPhone SE.

# Expected result

Page fits in screen width, no horizontal scroll.

# Actual result

You can scroll horizontally to the right and see blank bit of page:

<details>
<summary>Screenshots</summary>

<table>
<tr><td>
<img src=https://github.com/python/python-docs-theme/assets/1324225/17eebee9-c86c-4ae0-820b-31cd6c07cc9a>
<td>
<img src=https://github.com/python/python-docs-theme/assets/1324225/a78d2b5b-9679-47b5-b725-44d04e2e0990>
</table>

</details>

This is caused by long code literals that don't wrap:

https://docs.python.org/3.13/whatsnew/3.12.html#asyncio

<details>
<summary>Screenshots</summary>

<table>
<tr>
 <td>
<img src=https://github.com/python/python-docs-theme/assets/1324225/9d2b96f8-20ec-4452-80ec-733c5449717d>
 <td>
<img src=https://github.com/python/python-docs-theme/assets/1324225/0afcd3de-6891-4b40-aec5-b81436889cdb>
</table>

</details>

# Possible fix

Adding `overflow-wrap: break-word` to `span.pre` and changing `white-space: nowrap` to something else (e.g. `unset`) can help:

<details>
<summary>Screenshots</summary>

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td><img src=https://github.com/python/python-docs-theme/assets/1324225/6db800f5-4b7f-493c-8698-ee66e22a67df>

 <td><img src=https://github.com/python/python-docs-theme/assets/1324225/6905de40-6368-4275-a340-015e99b85d9b>

</table>

</details>

Re: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text

This `white-space: nowrap` CSS comes from https://docs.python.org/3.13/_static/basic.css which is from Sphinx:

https://github.com/sphinx-doc/sphinx/blob/d3c91f951255c6729a53e38c895ddc0af036b5b9/sphinx/themes/basic/static/basic.css_t#L762-L768

Should we fix it in Sphinx? 

[...]

PR #139, opened 2023-07-27
Don't let long code literals extend beyond the right side of the screen

hugovk:
Fixes https://github.com/python/python-docs-theme/issues/129.

With mobile, compare:

* https://docs.python.org/3.13/whatsnew/3.12.html
* https://python-docs-theme-previews--139.org.readthedocs.build/en/139/whatsnew/3.12.html

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td width=50% valign=top>

 <td width=50% valign=top>

<tr>
 <td>

 <td>

</table>

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -368,6 +368,11 @@ div.genindex-jumpbox a {
     body {
         margin-top: 40px;
     }
+    /* Don't let long code literals extend beyond the right side of the screen */
+    span.pre {
+        overflow-wrap: break-word;
+        white-space: unset;
+    }
 
     /* Top navigation bar */
     .mobile-nav {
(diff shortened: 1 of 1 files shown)


AA-Turner (approved):

pradyunsg:
In Furo, I went ahead and set break-word on the entire content container in https://github.com/pradyunsg/furo/commit/c8b51d09af3dcaac3046f7e761119e9d1b7c9e37

That makes it so that nothing, including really long one-word titles don't trigger an overflow of the content.

Do you think it's worth adopting a similar model here?

hugovk:
Yes, that sounds better, would you like to update this PR or create a new one?

hugovk:
Updated, how does this look?

pradyunsg (approved):
LGTM!

Issue #130, opened 2023-06-03
Focus search box when pressing slash


hugovk:
Many websites focus the search box when the forward slash <kbd>/</kbd> key is pressed.

For example:

* Furo theme https://pradyunsg.me/furo/quickstart/
* PyPI https://pypi.org/project/python-docs-theme/
* GitHub https://github.com/python/python-docs-theme/issues/130
* YouTube https://www.youtube.com/watch?v=nMekFX2CDVk
* Discord https://discuss.python.org/

Here's how PyPI implemented it: https://github.com/pypi/warehouse/pull/10395

hugovk:
Re-opening because the fix in #141 caused https://github.com/python/python-docs-theme/issues/144, and needs reverting (https://github.com/python/python-docs-theme/pull/146).

hugovk:
Fixed by https://github.com/python/python-docs-theme/pull/153.

PR #141, opened 2023-07-27
Use correct block, enable built-in slash keypress

miketheman:
Includes a commit removing a duplicate block.

When the template was made responsive, the choice was made to override
the `body_tag` fragment.
The `body_tag` block is a literal `<body>` which allows the theme developer
to set properties on the `body` tag.
See: https://github.com/sphinx-doc/sphinx/pull/4140

Instead, use the `document` block as documented.
https://www.sphinx-doc.org/en/master/development/templating.html#blocks

Overriding the correct block places the `mobile-nav` section _after_ the
majority of the other HTML parts, which enables the correct function of
the `/` keyboard shortcut.

Resolves https://github.com/python/python-docs-theme/issues/130
Closes https://github.com/python/python-docs-theme/pull/131
Closes https://github.com/python/python-docs-theme/pull/135

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -82,7 +82,7 @@
     {{ super() }}
 {%- endblock -%}
 
-{%- block body_tag %}
+{%- block document %}
 {{ super() }}
 {%- if builder != 'htmlhelp' %}
 <div class="mobile-nav">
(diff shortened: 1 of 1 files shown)


miketheman:
@hugovk I think this one is the thing we want!

hugovk:
Thanks!

Preview: https://python-docs-theme-previews--141.org.readthedocs.build/en/141/

With this, slash highlights the search box in the top right in desktop view:

image

But not the one in mobile view:

image

Which is the more useful way around.

hugovk (approved):

PR #157, opened 2023-10-04
Add changelog for 2023.9

hugovk:


hugovk:
Obviously, it's October and this should have been 2023.10... But then we did release 2023.7 in August and 2023.8 in September! Let's get back on track next time :)

PR #160, opened 2023-10-18
Underline links for readability and a11y

hugovk:
Currently this theme changes the colour of hyperlinks but doesn't underline them.

For example: https://docs.python.org/3/whatsnew/3.12.html

image

When hovering over a link, it changes colour and a link is shown (see "f-strings" here):

image

---

However, links should always be underlined. This improves accessibility and also readability.

For example:

* https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/links/do-not-remove-the-underline-from-links/
* https://adrianroselli.com/2019/01/underlines-are-beautiful.html

Not using underlines also fails a WCAG check:

* https://www.w3.org/TR/WCAG20-TECHS/F73.html

And GitHub have just enabled underlines:

* https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility/

# Demo

https://python-docs-theme-previews--160.org.readthedocs.build/en/160/whatsnew/3.12.html

image

image

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -30,7 +30,8 @@ pre {
 }
 
 a {
-    text-decoration: underline;
+    text-decoration: underline 1px;
+    text-underline-offset: 0.3em;
 }
 
 body {
@@ -181,7 +182,7 @@ div.body {
 
 div.body p, div.body dd, div.body li, div.body blockquote {
     text-align: left;
-    line-height: 1.4;
+    line-height: 1.6;
 }
 div.body h1, div.body h2, div.body h3, div.body h4, div.body h5, div.body h6 {
     margin: 0;
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -29,6 +29,10 @@ pre {
     color: inherit;
 }
 
+a {
+    text-decoration: underline;
+}
+
 body {
     margin-left: 1em;
     margin-right: 1em;
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -29,7 +29,7 @@ pre {
     color: inherit;
 }
 
-a {
+a[href] {
     text-decoration: underline 1px;
     text-underline-offset: 0.3em;
 }


ncoghlan (approved):

ncoghlan:
@hugovk I was about to hit merge on this, but then I wondered: Should we tag it for backport to the active maintenance branches as well?

Personally, I'm not seeing a downside to doing that, and a whole lot of benefits.

hugovk:
Yes, using this on the active maintenance branches is good. 

Note we're in the https://github.com/python/python-docs-theme repo, not https://github.com/python/cpython. We'll release this theme to PyPI, and then it'll be used by the backport branches the next time they rebuild:

* `main`: https://github.com/python/cpython/blob/e9b5399bee7106beeeb38a45cfef3f0ed3fdd703/Doc/requirements.txt#L18
* `3.12`: https://github.com/python/cpython/blob/68a5640a362be4f787f3da0e76cd174f7a458a4b/Doc/requirements.txt#L17
* `3.11`: https://github.com/python/cpython/blob/26a028269b807d599ea2f5fa266fccfd1421775b/Doc/requirements.txt#L17

ezio-melotti:
I find the screenshot with underlines "noisier" than the current one, and I suspect the lack of underlines was not added in the first place for similar reasons.  Wikipedia also doesn't have underlines and cites possible confusion when a link contains underscores (it might have a setting to still enable underlines though).

Given that both our docs and Wikipedia are rich in links, and that -- especially for our docs -- there are often underscores in the text of the links, do the guidelines that suggest adding underlines still apply?

For reference, here are a couple of examples of underlined links that contain underscores:
image
image

Perhaps the contrast could be tweaked instead, or a lighter/different (e.g. dotted) underline could be used instead?

hugovk:
Good point, let's consider narrowing/lowering the underline, and see how GitHub and MDN do it.

hugovk:
Updated to make the underline thinner, move it down a bit, and also increase the line height so it better fits in the gap. 

"Line height (line spacing) to at least 1.5 times the font size" is also one of the WCAG 2.1 guidelines, and improves accessibility through readability. 

https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html

Here's how the earlier snippets with underscores look:

image

image

And from What's New in 3.13:

image

How does this look?

ezio-melotti:
This is definitely better and solves the potential confusion with underscores.  I'm now -0 (from -1) on this since I still prefer the version without underlines, but if others prefer it I'm fine with the change.

python_docs_theme/static/pydoctheme.css
```diff
@@ -29,6 +29,11 @@ pre {
     color: inherit;
 }
 
+a {
+    text-decoration: underline 1px;
```
ezio-melotti, line 33:
```suggestion
    text-decoration: underline dotted 1px;
```
Have you tried this?

hugovk: <details>
<summary>Lots of screenshots</summary>


image

And if dotted we could reduce the `text-underline-offset` from 0.3em. Here's 0.2em:

image

0.1em:

image

Unspecified/default=auto:

image



</details>

hugovk: However, dotted underlines are sometimes reserved for "tooltips", as used on the PEPs:

* https://peps.python.org/
* https://github.com/python/peps/pull/2838

Another options is to have a lighter underline colour, like https://peps.python.org/ does. This does make the underline harder to see, and we need to be careful not to defeat the purpose of adding underlines in the first place.

python_docs_theme/static/pydoctheme.css
```diff
@@ -29,6 +29,11 @@ pre {
     color: inherit;
 }
 
+a {
```
pradyunsg, line 32:
```suggestion
a:any-link {
```

OR

```suggestion
a[href] {
```

Let's only underline actual links rather than [placeholder links](https://www.enovate.co.uk/blog/2013/07/16/html5-placeholder-links).

https://developer.mozilla.org/en-US/docs/Web/CSS/:any-link

pradyunsg (approved):

egeakman:
Hi there @hugovk. Is there Discourse discussion for this? I believe there should be a discussion first, I would like to hear people's opinions.

hugovk:
There wasn't a discussion for this, just this PR that was open for two months before merge, and one month before release.

This was prompted by GitHub recently [adding underlines for accessibility](https://github.blog/changelog/2023-10-18-new-default-underlined-links-for-improved-accessibility/). In addition to approvals here, core devs were in support (on Discord), and some said underlines help with poor eyesight or colourblindness.

Some references:

* https://www.w3.org/TR/WCAG20-TECHS/F73.html
* https://adrianroselli.com/2016/06/on-link-underlines.html
* https://www.nngroup.com/articles/guidelines-for-visualizing-links/

PR #165, opened 2024-01-26
Add `hosted_on` variable for a link in the footer

hugovk:
Helps https://github.com/python/docs-community/issues/5.

If `hosted_on` is defined in your docs' `conf.py`:

```python
html_theme_options = {
    'collapsiblesidebar': True,
    'hosted_on': '<a href="https://example.com">Example</a>',
    'issues_url': '/bugs.html',
    'license_url': '/license.html',
    'root_include_title': False   # We use the version switcher instead.
}
```

Adds a "Hosted by" line to the footer:

image

# Conditional

For https://github.com/python/docs-community/issues/5, we can instead add this in `conf.py`:

```python
if os.getenv("READTHEDOCS"):
    html_theme_options["hosted_on"] = '<a href="https://about.readthedocs.com/">Read the Docs</a>'
```

And we'll only get the "Hosted by" line for [builds created on Read the Docs](https://docs.readthedocs.io/en/stable/reference/environment-variables.html):

image

Here's a demo build:

* https://hugovk-cpython.readthedocs.io/en/hosted-on/
* https://github.com/hugovk/cpython/tree/hosted-on

If you build that locally or on another server which doesn't have the env var defined, we don't get the link:

image

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -143,6 +143,9 @@
     {% trans %}Examples, recipes, and other code in the documentation are additionally licensed under the Zero Clause BSD License.{% endtrans %}
     <br />
     {% if theme_license_url %}{% trans license_file=theme_license_url %}See <a href="{{ license_file }}">History and License</a> for more information.{% endtrans %}<br />{% endif %}
+    {% if theme_hosted_on %}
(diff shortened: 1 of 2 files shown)


willingc (approved):
Looks good to me. Thanks @hugovk

hugovk:
This has now been released, so here's the follow-up: https://github.com/python/cpython/pull/114697

Issue #168, opened 2024-01-31
Disable underlined links


denballakh:
Most of the web does not underline links by default. Underlined links look noisy and ugly.
Please disable them or at least provide an option to disable them.

This PR #160 mentioned that GitHub enabled them. But what you did is hugely different from what GitHub did: you force users to look at underlines instead of giving them the option to do so.

mikeshardmind:
https://www.a11y-101.com/development/underlined-links Just going to leave this here. This is a significantly better default for those who actually need it rather than those who just think it is noisy or ugly.

wookie184:
I don't find the underlines noisy or ugly in general text. I do think that the underlines make underscores harder to distinguish from spaces though, which isn't ideal.

The PR linked mentioned that not using underlines also fails [WCAG check F73](https://www.w3.org/TR/WCAG20-TECHS/F73.html). This isn't quite correct, the check says "Link underlines or some other non-color visual distinction are required". You can use https://webaim.org/resources/linkcontrastchecker/ to check if a link has sufficient contrast with normal text to pass the check - currently the light theme passes but not the dark theme, but that wouldn't be hard to fix.

Another question is whether to have different behaviour in the table of contents, the WCAG check also mentions "While some links may be visually evident from page design and context", which I would say is the case for a table of contents, and that is the one place where I think the noisiness argument is fair - having all the text underlined does make it slightly harder to read for me.

All those points made, I don't feel strongly either way and would personally be fine with how it is now.

ezio-melotti:
For reference:
* underlines were added in #160, which also contains a discussion with several references
* the distance from the text/underscores was then further tweaked in #166

A setting to enable/disable underlines (like GitHub did) can always be added.  This is a bit more involved than simply changing the default by tweaking a couple of lines of CSS though, so for the first iteration we opted to keep it simple.  For the next iteration we just need to agree on how/where to put the switch, and have someone contribute a PR.

FWIW I also prefer non-underlined links, but I can live with underlines as a default if they are more accessible.

quicknir:
Since we already have a toggle for light/dark (which presumably works via cookie or something), I think that a toggle for underlines is definitely the "best" solution but it's obviously also the most work. It's not a ton of work perhaps but it is more than 0. I guess it boils down to whether someone is actually capable and willing willing to do the work; if it is then I think it's a pretty easy win but if not then it doesn't really matter if it's the best solution or not.

mikeshardmind:
[...]

PR #169, opened 2024-02-02
Do not underline navigation links

hugovk:
Follow on from https://github.com/python/python-docs-theme/pull/160.

Fixes https://github.com/python/python-docs-theme/issues/168.

# Accessibility

The lowest level WCAG* 2.0 guidelines (specifically [G182](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20160317/G182)) say colour alone shouldn't be used for links. Something else -- like an underline, different font or size, or italics/bold -- must also be used to help identify them.

However, this is mainly for main body text (e.g. words linked in a paragraph or narrative content), not thing like navigation, footers, page controls.

> ## Applicability
> Colored text when the color is used to convey information such as:
> * Words that are links in a paragraph
> * Items in a list where some are different than others and are presented in colored text
> ## Description
> The intent of this technique is to provide a redundant visual cue for users who may not be able to discern a difference in text color. Color is commonly used to indicate the different status of words that are part of a paragraph or other block of text or where special characters or graphics cannot be used to indicate which words have special status. For example, scattered words in text may be hypertext links that are marked as such by being printed in a different color. This technique describes a way to provide cues in addition to color so that users who may have difficulty perceiving color differences or have low vision can identify them.
> 
> To use this technique, an author incorporates a visual cue in addition to color for each place where color alone is used to convey information. Visual cues can take many forms including changes to the font style, the addition of underlines, bold, or italics, or changes to the font size.

(*There are many international accessibility laws based on WCAG: https://www.w3.org/WAI/policies/)

# Usability

[Usability guidelines](https://www.nngroup.com/articles/guidelines-for-visualizing-links/) suggest they can be removed from navigation (but if accessibility is not a priority):

> * To maximize the [perceived affordance](https://jnd.org/affordances_and_design/) of clickability, color and underline the link text. Users shouldn't have to guess or scrub the page to find out where they can click.
> * Assuming the link text is colored, it's not always absolutely necessary to underline it.
> 
>   * There are two main cases in which you can safely eliminate underlines: navigation menus and other lists of links. However, this is true only when the page design clearly indicates the area's function. (Remember: your design might not be as obvious to outside users as it is to your own team members.) Users typically understand a left-hand navigation rail with a list of links on a colored background, assuming it resembles the navigation areas on most other sites.
>   * Exception: underlining is essential if you use link colors such as reds or greens, which cause problems for users with common forms of color-blindness.
>   * Exception: underlined links are important for low-vision users' [accessibility](http://www.nngroup.com/topic/accessibility/), so retain underlines if accessibility is a priority for your site or you have many users with low vision.

# This PR

Disables underlines for:

* header navigation
* footer navigation
* left menu bar
* list of links on the homepage
* list of links in the global module index
* list of links in the general module index
* list of links in search results
* tables of contents

Excepts re-enables the underlines when mouse hovers over them (also the pre-#160 behaviour).

Also, use the default underline offset in the footer. We had links there pre-#160 too.

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -33,6 +33,30 @@ a[href] {
     text-decoration: underline 1px;
     text-underline-offset: 0.25em;
 }
+/* No underline for navigation */
+div.genindex-jumpbox a,
+div.modindex-jumpbox a,
+div#search-results a,
+div.sphinxsidebar a,
+div.toctree-wrapper a,
+div[role=navigation] a,
+table.contentstable a,
+table.indextable a {
+    text-decoration: none;
+}
+
+/* Except when hovered */
+div.genindex-jumpbox a:hover,
+div.modindex-jumpbox a:hover,
+div#search-results a:hover,
+div.sphinxsidebar a:hover,
+div.toctree-wrapper a:hover,
+div[role=navigation] a:hover,
+table.contentstable a:hover,
+table.indextable a:hover {
+    text-decoration: underline;
+    text-underline-offset: auto;
+}
 
 body {
     margin-left: 1em;
@@ -282,6 +306,10 @@ div.footer {
     text-align: right;
     width: auto;
     margin-right: 10px;
+}
+
+div.footer a {
+    text-underline-offset: auto;
 }
 
 div.footer a:hover {


python_docs_theme/static/pydoctheme.css
```diff
@@ -33,6 +33,31 @@ a[href] {
...
+div.genindex-jumpbox a:hover,
+div.modindex-jumpbox a:hover,
+div#search-results a:hover,
+div.sphinxsidebar a:hover,
+div.toctree-wrapper a:hover,
+div[role=navigation] a:hover,
+table.contentstable a:hover,
+table.indextable a:hover {
+    text-decoration: underline;
+    text-underline-offset: auto;
```
ezio-melotti, line 59: `auto` should already be the default -- is there something else overriding this when it's not explicitly stated, or did you add it as a precaution to prevent other rules from overriding the value in case the offset is added to some parent container in the future?

hugovk: > is there something else overriding this when it's not explicitly stated ...

Yes, it's overridden by this:

```css
a[href] {
    text-decoration: underline 1px;
    text-underline-offset: 0.25em;
}
```

So "normal" body text links have underlines with extra offset; but navigational links only have underlines for hover, and they're the default offset.

ezio-melotti (approved):

PR #170, opened 2024-02-02
Update config

hugovk:
* Add Dependabot to keep GitHub Actions up-to-date
* Update pre-commit, add more handy checks, remove some that don't apply
* Move isort config to pyproject.toml and specify Black profile for compatibility
* Add link to PR of RTD docs preview

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: monthly
+    assignees:
+      - "ezio-melotti"
     groups:
       actions:
         patterns:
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly
```
ezio-melotti, line 6:
```suggestion
    schedule:
      interval: monthly
    assignees:
      - "ezio-melotti"
```

I don't mind being pinged about these, as I already handle them on a few different repos.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+    groups:
+      actions:
+        patterns:
+          - "*"
```
ezio-melotti, line 10: Any reason for the grouping?  Sometimes deps are related and must be updated together, but other times one update is problematic and it will block the others.

hugovk: Mainly to reduce the total number of PRs to reduce "alert fatigue". I and many others get lots of PR/issue notifications, and these multiply on Dependabot days.

Other projects I maintain use grouped updates, and very occasionally one update may be problematic, but like you say, it goes both ways and sometimes they _need_ grouping. 

The vast majority of the time, grouped updates are fine.

Multiple PRs use up more CI time, and are also more cumbersome: after merging one, often it needs to rebase all the others, using more CI time, and needing to wait before merge.

PR #166, opened 2024-01-27
Reduce underline offset

hugovk:
We added underlines in https://github.com/python/python-docs-theme/pull/160, with an offset to make sure they don't obscure underscores in code formatting. With fresh eyes, 0.3em looks a bit too much. Let's reduce it.

# 0.3em

image

# 0.2em

image

PR #171, opened 2024-02-03
Only apply underline offset to code formatting for underline visibility

hugovk:
Helps https://github.com/python/python-docs-theme/issues/168.

In #160 we added underlines to links, with an extra offset to help make sure underscores (`_`) are not obscured by the underlines.

(In https://github.com/python/python-docs-theme/pull/166 we reduced the amount of offset.)

We only really need the offset for code formatting, where underscores often show up, so we can use the default offset for normal links and only apply an extra offset for code.


ezio-melotti (approved):
I would add a comment explaining what these two rules do, otherwise LGTM.

PR #176, opened 2024-03-16
Modernise font: use system font stack to improve text readability and webpage performance

hugovk:
The primary font used for by this theme is [Lucida Grande](https://en.wikipedia.org/wiki/Lucida_Grande).

By default, this is only used on macOS:

<details>
<summary>macOS</summary>

macOS

</details>

On other platforms, the browser uses the more common Arial:

<details>
<summary>Windows, Linux, Android, iPhone</summary>

| Windows | Linux | Android | iPhone |
|  - | - | - | - |
| Windows | Linux | Android | iPhone |

</details>

# History

Lucida Grande used to be the system font on Macs from 1999 to 2014, and was used by websites such as apple.com, facebook.com and twitter.com.

* In OS X Yosemite 10.10 (released in 2014), Apple changed the system font to [Helvetica Neue](https://en.wikipedia.org/wiki/Helvetica#Neue_Helvetica_(1983)) (much closer to Arial).
* In OS X El Capitan 10.11 (released in 2015), Apple changed it to [San Francisco](https://en.wikipedia.org/wiki/San_Francisco_(sans-serif_typeface)).

I can't think of any other big websites still using Lucida Grande. To me it looks dated, and I find it hard to read.

# System font stack

I recommend we update to the so-called "system font stack".

Benefits according to https://systemfontstack.com:

* Fast: No network request, no time to parse a font, no flash of an incorrect font.
* Styles & unicode: System fonts have lots of styles and broad language coverage, unlike many webfonts.
* Familiarity: Web apps feel more native when they use system font faces.

Many sites use a system font stack: 
* [GitHub](https://markdotto.com/2018/02/07/github-system-fonts/)
* [Stack Overflow](https://meta.stackexchange.com/q/364048/162511)
* [Booking.com](https://booking.design/implementing-system-fonts-on-booking-com-a-lesson-learned-bdc984df627f)
* [Weather.com, Bootstrap, Medium, Ghost, PubMed, and the WordPress dashboard](https://woorkup.com/system-font/)
* [Furo](https://github.com/pradyunsg/furo/blob/3c4e94897b764a0e25bf6bcf7dd74356009918fe/src/furo/assets/styles/variables/_fonts.scss#L8) theme (used on the [devguide](https://devguide.python.org/))
* [PyData Sphinx Theme](https://github.com/pydata/pydata-sphinx-theme/pull/285#issuecomment-730571293)
* https://peps.python.org/: [python/peps#3132](https://github.com/python/peps/pull/3132)

See more on the [history and rationale](https://medium.com/web-dev-survey-from-kyoto/survey-system-font-stack-5f73a3b39776).

# Preview

<details>
<summary>macOS</summary>

image

</details>

https://python-docs-theme-previews--176.org.readthedocs.build/en/176/

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -170,7 +170,7 @@ div.sphinxsidebar a:hover {
 form.inline-search input,
 div.sphinxsidebar input,
 div.related input {
-    font-family: 'Lucida Grande', Arial, sans-serif;
+    font-family: -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif;
     border: 1px solid #999999;
     font-size: smaller;
     border-radius: 3px;
@@ -258,7 +258,7 @@ div.body a:hover {
 }
 
 tt, code, pre {
-    font-family: ui-monospace, "Cascadia Mono", "Segoe UI Mono", "Liberation Mono", Menlo, Monaco, Consolas, monospace;
+    font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
     font-size: 96.5%;
 }
 
@@ -362,7 +362,7 @@ div.genindex-jumpbox a {
     top: 0;
     right: 0;
     text-size: 75%;
-    font-family: monospace;
+    font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
     padding-left: 0.2em;
     padding-right: 0.2em;
     border-radius: 0 3px 0 0;
--- a/python_docs_theme/theme.conf
+++ b/python_docs_theme/theme.conf
@@ -5,8 +5,8 @@ pygments_style = default
 pygments_dark_style = monokai
 
 [options]
-bodyfont = 'Lucida Grande', Arial, sans-serif
-headfont = 'Lucida Grande', Arial, sans-serif
+bodyfont = -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif
+headfont = -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif
 footerbgcolor = white
 footertextcolor = #555555
 relbarbgcolor = white


hugovk:
The 3.8 and 3.9 failures unrelated, see https://github.com/python/python-docs-theme/pull/177 for a fix.

ronaldoussoren:
Apple supports https://drafts.csswg.org/css-fonts-4/#ui-serif-def for this, with 'ui-serif', 'ui-sans-serif' and 'ui-monospace'.   The last one allows using the system monospace font (SF Mono) without explicitly naming it.  

These aren't supported by other browser at this time (according to caniuse)

hugovk:
Thanks, I didn't know about those! The caniuse link: https://caniuse.com/extended-system-fonts, shows 17.33% availability.

The resources tab links to the [Chromium](https://issues.chromium.org/issues/40661155) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1598880) issues, but there's not been movement for 1-2 years. [Webkit](https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/) (Safari) has supported it for 4 years.

The "CSS Fonts Module Level 4" spec is still draft, but was updated last month.

We could put first in `font-family`, and non-Safari browsers would fall back to the other fonts. But perhaps we should play safe and not use them until more browsers support it or the spec is approved? Or are they already widely used by other sites?

willingc (approved):
I would really like to see the modernized font.

pradyunsg (approved):

willingc:
🚢 @hugovk Thanks!

JacobCoffee (approved):

pradyunsg:
> The "CSS Fonts Module Level 4" spec is still draft, but was updated last month.

I would prefer to wait until this spec was implemented in more browsers. If we go down this route, it's worth deciding if we would prefer to be stylistically consistent based on the OS + available fonts vs aligned with the user's OS font configuration (which is likely more varied on Linux and Android, rather than MacOS and Windows).

---

I suggest we take on things from the updated font stack in a follow up, since this is a meaningful improvement over status quo!

Privat33r-dev:
Looks great, thanks for the PR

ronaldoussoren:
> Thanks, I didn't know about those! The caniuse link: https://caniuse.com/extended-system-fonts, shows 17.33% availability.
> 
> The resources tab links to the [Chromium](https://issues.chromium.org/issues/40661155) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1598880) issues, but there's not been movement for 1-2 years. [Webkit](https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/) (Safari) has supported it for 4 years.
> 
> The "CSS Fonts Module Level 4" spec is still draft, but was updated last month.
> 
> We could put first in `font-family`, and non-Safari browsers would fall back to the other fonts. But perhaps we should play safe and not use them until more browsers support it or the spec is approved? Or are they already widely used by other sites?

I don't know.

I was primarily triggered by the list of fonts for "copybutton", which includes Apple's previoius preferred monospace font (Menlo), but not the current one (SF Mono). Using "ui-monospace" there would future proof us a little for Apple's platforms, although I'd assume that they'll keep using the San Francisco font suite for the foreseeable future.

PR #178, opened 2024-03-17
Correct invalid property name

Privat33r-dev:


pradyunsg (approved):

hugovk:
Yes, we should fix this :)

But should we fix it by _removing_ `text-size`, and not reducing the font size? So essentially maintain the status quo? I think I prefer the bigger ">>>" button:

# Status quo: not setting `font-size`

image

https://python-docs-theme-previews.readthedocs.io/en/latest/tutorial/introduction.html#numbers

# PR: `font-size: 75%;`

image

https://python-docs-theme-previews--178.org.readthedocs.build/en/178/tutorial/introduction.html#numbers

Privat33r-dev:
From one point, it fixes the CSS, from the other point, it makes the element a bit small
image
image

As I was writing, you posted the same thing as well. I think that removal of the `text-size` property would be more desirable and if 2 people have the same idea independently, I guess that it's a valid point.

hugovk (approved):
👍 

@pradyunsg What do you think?

ezio-melotti (approved):

pradyunsg (approved):

pradyunsg:
> @pradyunsg What do you think?

Good catch, and looks like this is good to merge! :)

PR #188, opened 2024-04-17
Use different colour for 'Return value: Borrowed reference'

hugovk:
Follow on from https://github.com/python/cpython/pull/117926.

Currently light green is used for all return types:

* "Return value: Always NULL."
* "Return value: New reference."
* "Return value: Borrowed reference."

For example: https://docs.python.org/3/c-api/exceptions.html

image

image

image

A borrowed reference is not necessarily a "bad" thing, but we may want to use something other than green that indicates it's "good", to indicate they should be treated with care.

This PR uses the same yellow/orange colour as the "Changed in version x.y" text in https://github.com/python/python-docs-theme/pull/185 - I've already checked the contrast ratios in that PR and both meet WCAG AAA. By using the same colour, it will be easier to check/change when we audit.

image

image

For example: https://python-docs-theme-previews--188.org.readthedocs.build/en/188/c-api/exceptions.html#querying-the-error-indicator

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -1,5 +1,15 @@
 @import url('classic.css');
 
+/* Common colours */
+:root {
+    --good-color: rgb(41 100 51);
+    --good-border: rgb(79 196 100);
+    --middle-color: rgb(133 72 38);
+    --middle-border: rgb(244, 227, 76);
+    --bad-color: rgb(159 49 51);
+    --bad-border: rgb(244, 76, 78);
+}
+
 /* unset some styles from the classic stylesheet */
 div.document,
 div.body,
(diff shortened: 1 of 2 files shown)


vstinner:
> This PR uses the same yellow/orange colour as the "Changed in version x.y" text in https://github.com/python/python-docs-theme/pull/185

FYI As a color blind person, I don't see the difference between "Borrowed reference" and "New reference" colors in the **light mode**. Well, I'm also color blind, so my opinion may not matter :-)

In the dark mode, I see clearly that the two colors are very different.

flying-sheep:
If you’re reusing existing colors, how about you centralize their definition as a CSS variables and then use it from both your new CSS rules and the old ones that originally defined the color?

erlend-aasland (approved):
I don't have a strong opinion regarding naming, but I like the change! Personally, I would prefer borrowed references to be labelled as "bad", but that is a different discussion.

hugovk:
This is now deployed: https://docs.python.org/3/c-api/exceptions.html

PR #190, opened 2024-06-09
Add backgrounds and borders to admonitions

hugovk:
Only some of the "admonitions" have backgrounds and borders, giving them slightly different indentations:

image

We're missing backgrounds and borders for most of them:

* https://sphinx-themes.org/sample-sites/python-docs-theme/kitchen-sink/admonitions/

Other themes have them for all:

* https://sphinx-themes.org/sample-sites/pydata-sphinx-theme/kitchen-sink/admonitions/
* https://sphinx-themes.org/sample-sites/furo/kitchen-sink/admonitions/
* https://sphinx-themes.org/sample-sites/sphinx-rtd-theme/kitchen-sink/admonitions/
* https://sphinx-themes.org/sample-sites/default-alabaster/kitchen-sink/admonitions/

Let's also fill in the gaps.

# Before

<table>
<tr>
<td>image
<td>image
</table>

# After

<table>
<tr>
<td>image
<td>image
</table>

Screenshots using this from the [Kitchen Sink](https://github.com/sphinx-themes/sphinx-themes.org/blob/master/sample-docs/kitchen-sink/admonitions.rst?plain=1):

<details>
<summary>Details</summary>

```rst
..
   Copyright (c) 2021 Pradyun Gedam
   Licensed under Creative Commons Attribution-ShareAlike 4.0 International License
   SPDX-License-Identifier: CC-BY-SA-4.0

===========
Admonitions
===========

Sphinx provides several different types of admonitions.

``topic``
=========

.. topic:: This is a topic.

   This is what admonitions are a special case of, according to the docutils
   documentation.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``admonition``
==============

.. admonition:: The one with the custom titles

   It's got a certain charm to it.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``attention``
=============

.. attention::

   Climate change is real.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``caution``
===========

.. caution::

   Cliff ahead: Don't drive off it.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``danger``
==========

.. danger::

   Mad scientist at work!

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``error``
=========

.. error::

   Does not compute.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``hint``
========

.. hint::

   Insulators insulate, until they are subject to ______ voltage.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``important``
=============

.. important::

   Tech is not neutral, nor is it apolitical.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``note``
========

.. note::

   This is a note.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``seealso``
===========

.. seealso::

   Other relevant information.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``tip``
=======

.. tip::

   25% if the service is good.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.

``warning``
===========

.. warning::

   Reader discretion is strongly advised.

   We also support *italic*, **bold**, ``code``, `links <https://www.sphinx-doc.org/>`_, and more.
```

</details>

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -250,7 +250,7 @@ div.body pre {
     --attention-background: #bbddff5c;
     --attention-border: #0000ff36;
     --caution-background: #ffc;
-    --caution-border: #dddd66;
+    --caution-border: #dd6;
     --danger-background: #ffe4e4;
     --danger-border: red;
     --error-background: #ffe4e4;
@@ -258,7 +258,7 @@ div.body pre {
     --hint-background: #dfd;
     --hint-border: green;
     --seealso-background: #ffc;
-    --seealso-border: #dddd66;
+    --seealso-border: #dd6;
     --tip-background: #dfd;
     --tip-border: green;
     --warning-background: #ffe4e4;
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -242,17 +242,80 @@ div.body pre {
     border: 1px solid #ac9;
 }
 
-div.body div.admonition,
+/* Admonitions */
+:root {
+    --admonition-background: #eee;
+    --admonition-border: #ccc;
+    --attention-background: #bbddff5c;
+    --attention-border: #0000ff36;
+    --caution-background: #ffc;
+    --caution-border: #dddd66;
+    --danger-background: #ffe4e4;
+    --danger-border: red;
+    --error-background: #ffe4e4;
+    --error-border: red;
+    --hint-background: #bfc;
+    --hint-border: green;
+    --seealso-background: #ffc;
+    --seealso-border: #dddd66;
+    --tip-background: #bfc;
+    --tip-border: green;
+    --warning-background: #ffe4e4;
+    --warning-border: red;
+}
+
+div.body div.admonition {
+    background-color: var(--admonition-background);
+    border-radius: 3px;
+    border: 1px solid var(--admonition-border);
+}
+
+div.body div.admonition.attention {
+    background-color: var(--attention-background);
+    border-color: var(--attention-border);
+}
+
+div.body div.admonition.caution {
+    background-color: var(--caution-background);
+    border-color: var(--caution-border);
+}
+
+div.body div.admonition.danger {
+    background-color: var(--danger-background);
+    border-color: var(--danger-border);
+}
+
+div.body div.admonition.error {
+    background-color: var(--error-background);
+    border-color: var(--error-border);
+}
+
+div.body div.admonition.hint {
+    background-color: var(--hint-background);
+    border-color: var(--hint-border);
+}
+
+div.body div.admonition.seealso {
+    background-color: var(--seealso-background);
+    border-color: var(--seealso-border);
+}
+
+div.body div.admonition.tip {
+    background-color: var(--tip-background);
+    border-color: var(--tip-border);
+}
+
+div.body div.admonition.warning {
+    background-color: var(--warning-background);
+    border-color: var(--warning-border);
+}
+
 div.body div.impl-detail {
     border-radius: 3px;
 }
 
 div.body div.impl-detail > p {
     margin: 0;
-}
-
-div.body div.seealso {
-    border: 1px solid #dddd66;
 }
 
 div.body a {
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -246,6 +246,7 @@ div.body pre {
 :root {
     --admonition-background: #eee;
     --admonition-border: #ccc;
+    --admonition-color: black;
     --attention-background: #bbddff5c;
     --attention-border: #0000ff36;
     --caution-background: #ffc;
@@ -268,6 +269,7 @@ div.body div.admonition {
     background-color: var(--admonition-background);
     border: 1px solid var(--admonition-border);
     border-radius: 3px;
+    color: var(--admonition-color);
 }
 
 div.body div.admonition.attention {
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -254,11 +254,11 @@ div.body pre {
     --danger-border: red;
     --error-background: #ffe4e4;
     --error-border: red;
-    --hint-background: #bfc;
+    --hint-background: #dfd;
     --hint-border: green;
     --seealso-background: #ffc;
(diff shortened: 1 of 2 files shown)


python_docs_theme/static/pydoctheme.css
```diff
@@ -242,7 +242,74 @@ div.body pre {
...
+    --seealso-border: #dddd66;
+    --tip-background: #bfc;
+    --tip-border: green;
+    --warning-background: #ffe4e4;
+    --warning-border: red;
+}
+
+div.body div.admonition {
+    background-color: var(--admonition-background);
+    border-radius: 3px;
```
nedbat, line 269: The most minor of nits: border-radius should come first so the two color lines can be together as in the others.

python_docs_theme/static/pydoctheme.css
```diff
@@ -242,7 +242,74 @@ div.body pre {
...
+    --admonition-border: #ccc;
+    --attention-background: #bbddff5c;
+    --attention-border: #0000ff36;
+    --caution-background: #ffc;
+    --caution-border: #dddd66;
+    --danger-background: #ffe4e4;
+    --danger-border: red;
+    --error-background: #ffe4e4;
+    --error-border: red;
+    --hint-background: #bfc;
```
nedbat, line 257: `#bfc` is a bit vibrant for my tastes.  I'm also curious why attention has alpha specified?  Where did these colors come from?

hugovk: For new colours, I started with the ones in the PEPs repo. 

I then adjusted them using Chrome's dev tools to shift the colour a bit, but mostly the alpha, so the contrast was high enough to meet WCAG AA. It was mainly the hyperlink that needed the adjustment. 

Happy to change any of these. Any suggestions? Or I can try toning down the `#bfc` a bit?

We can also replace the blue and green with the default grey, if we want to fine tune colours in another PR. 

We might also want to replace some of the other default greys in followup PRs, compare some of the other themes' Kitchen Sink demos.

nedbat: I'd change `#bfc` to `#dfd`, it gets the point across.

hugovk: How's this?

image

nedbat: I like it. 👍

nedbat (approved):

python_docs_theme/static/pydoctheme.css
```diff
@@ -242,7 +242,74 @@ div.body pre {
...
+    --error-background: #ffe4e4;
+    --error-border: red;
+    --hint-background: #dfd;
+    --hint-border: green;
+    --seealso-background: #ffc;
+    --seealso-border: #dddd66;
+    --tip-background: #dfd;
+    --tip-border: green;
+    --warning-background: #ffe4e4;
+    --warning-border: red;
```
ezio-melotti, line 264: Would it be better to use a consistent notation (e.g. 6/8 hex digits) instead of mixing 3/6/8 hex and names?

python_docs_theme/static/pydoctheme.css
```diff
@@ -242,7 +242,74 @@ div.body pre {
...
+}
+
+div.body div.admonition {
+    background-color: var(--admonition-background);
```
ezio-melotti, line 268: It's usually good practice to specify both `color` and `background-color`.  If you rely on an inherited `color`, it might change and end up conflicting with the `background-color`.

python_docs_theme/static/pydoctheme_dark.css
```diff
@@ -113,6 +113,28 @@ div.warning {
...
+    --error-background: rgba(255, 0, 0, 0.2);
+    --error-border: #f66;
+    --hint-background: #0044117a;
+    --hint-border: green;
+    --seealso-background: rgba(255, 255, 0, 0.1);
+    --seealso-border: #dd6;
+    --tip-background: #0044117a;
+    --tip-border: green;
+    --warning-background: rgba(255, 0, 0, 0.2);
+    --warning-border: #f66;
```
ezio-melotti, line 135: These could be more consistent too, since it's a mix between names, 3/8 hex digits, and `rgba()`.

ezio-melotti (commented):
I left a few nitpicky comments, and I also have some additional bikeshedding, since the color choice of some admonition seems somewhat arbitrary to me:
* `attention` seems to belong to the category of "potentially dangerous things that you should pay attention too", alongside `important`, `caution`, `danger`, `warning`, and `error`.  I would probably place it between `important` and `caution`, color them gray/yellow, and then proceed with orange/red for more dangerous things -- using blue for `attention` feels wrong to me.
* Similarly `hint`, `note`, `see also`, `tip` are all more neutral admonitions, and for those those I would use a neutral color like gray, blue, or possibly yellow -- green stands out a bit too much IMHO.

We should also check how they look on some real-world example, and see if the colors (and maybe the boxes too) are too distracting.  This might actually be the reason why they weren't added in the first place (maybe we should check with the original author too and ask them why they made that choice).

We might also want to leave the bikeshedding for a separate PR (especially if you just copied the colors from another theme) as not to stall this PR.

python_docs_theme/static/pydoctheme.css
hugovk: Updated to replace rgba with hex values, and preferring shorter hex where available. Also keeping names where available because those are more explicit.

python_docs_theme/static/pydoctheme.css
hugovk: Added.

python_docs_theme/static/pydoctheme_dark.css
hugovk: Updated to replace rgba with hex values, and preferring shorter hex where available. Also keeping names where available because those are more explicit.

hugovk:
Thanks all for the reviews, let's merge and release. We can adjust and iterate the others in the future.

Issue #194, opened 2024-07-31
Let's localize this theme


rffontenelle:
While Python docs is available for translation, this theme is still used in English. I see strings marked for translation, but there is no localization infra (no message catalog template; no babel's extract, update, compile, etc. commands set; no translation platform) or workflow set.

Is this project accepting contributions for enabling the theme's localization?

AA-Turner:
> Is this project accepting contributions for enabling the theme's localization?

I think we should.

Do you have recommendations for what we ought to do?

A

rffontenelle:
Basic steps are:
- choose a workflow (receive translations and update source messages as a pre-release step during a string freeze, or automatically update source messages and receive translations anytime?)
- set up a script that runs babel's extract, init, update and compile options (via [pybabel](https://babel.pocoo.org/en/latest/cmdline.html) or customized Python script that calls babel)
- optionally set up a translation project in a web-based translation platform like Weblate or Transifex to gather contributions in a more translation-friendly environment (so the translator doesn't have to know how to handle git concepts or GitHub pull requests)

My 2 cents:
- host translations in Hosted Weblate (free of charge for free/libre software)
- have a CI workflow that updates POT file whenever there is a new commit; with [webhook set](https://docs.weblate.org/en/latest/admin/continuous.html#automatically-receiving-changes-from-github), Weblate will make available new strings for every language
- have a script for running the babel's commands; this script will be used by the CI workflow, and may be used manually by an maintainer (see Sphinx's [babel_runner.py](https://github.com/sphinx-doc/sphinx/blob/master/utils/babel_runner.py) and pydata's [noxfile.py](https://github.com/pydata/pydata-sphinx-theme/blob/main/noxfile.py)).

StanFromIreland:
Any progress? We have the script but it not receiving much use at the moment

rffontenelle:
My bad. I'll get back to this. Indeed there are not many strings, so it's an easy translation.

StanFromIreland:
No worries! Webplate/Transifex might just be overkill for this one. All translations have at least one translator that can submit PRs and I'd say this would under take an hour to translate.

m-aciek:
My opinion: let's use just PO files in the repository, yet without a platform. And move it to Weblate along the Python docs (in the same project, for better discoverability), when it will be set up (hopefully soon).

StanFromIreland:
When this is implemented, I am happy to write a section for the devguide on contributing here:-)

[...]

PR #195, opened 2024-08-01
Add script for handling translations

rffontenelle:
This is an initial take on the enable localization initiative adding a script to ease the babel call.

Usage:

```shell
python babel_runner.py extract           # generate translations template file (.pot)
python babel_runner.py init -l pt_BR     # generate a translation file (.po) for Brazilian Portuguese
python babel_runner.py update            # update all .po files using the existing .pot; or
python babel_runner.py update -l pt_BR   # update only the pt_BR.po
python babel_runner.py compile           # compile all .po files using the existing .pot; or
python babel_runner.py compile -l pt_BR  # compile only the pt_BR.po
```

See #194


babel_runner.py
```diff
@@ -0,0 +1,108 @@
...
+
+# Global variables used by pybabel below
+DOMAIN = "messages"
+COPYRIGHT_HOLDER = "Python Software Foundation"
+LOCALES_DIR = os.path.relpath(os.path.join(PROJECT_DIR, "locales"))
+POT_FILE = os.path.relpath(os.path.join(LOCALES_DIR, f"{DOMAIN}.pot"), PROJECT_DIR)
+SOURCE_DIR = os.path.relpath(
+    os.path.join(PROJECT_DIR, "python_docs_theme"), PROJECT_DIR
+)
+MAPPING_FILE = os.path.relpath(os.path.join(PROJECT_DIR, "babel.cfg"), PROJECT_DIR)
```
hugovk, line 19: Perhaps prefix the config file with a dot?
```suggestion
MAPPING_FILE = os.path.relpath(os.path.join(PROJECT_DIR, ".babel.cfg"), PROJECT_DIR)
```

hugovk (commented):
Thanks for this! Please could you add it somehow to the CI, so we know it at least runs without errors?

I was thinking if this script could be converted into `Makefile` commands, but then we won't have Windows support, so that's no good. But we could add Makefile commands to call this script. This can be left for later though.

babel_runner.py
rffontenelle: Working on a GitHub Actions workflow setup to test the translation files. I renamed the mapping file to `.babel.cfg`. By the way, I was about to go crazy with babel\<tab> completing with both mapping file and script instead of only the script, so thanks for the suggestion!

rffontenelle:
@hugovk workflow job added to tests.yml

.github/workflows/tests.yml
```diff
@@ -48,3 +48,36 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          pip install -U pip setuptools
+          pip install Babel jinja2
+      - run: python3 babel_runner.py extract
+      - run: python3 babel_runner.py init -l pt_BR
+      - run: python3 babel_runner.py update
+      - run: python3 babel_runner.py update -l pt_BR
+      - run: python3 babel_runner.py compile
+      - run: python3 babel_runner.py compile -l pt_BR
```
Wulian233, line 74: Here you are using python3 and using 

> os: ["ubuntu-latest", "windows-latest"] 

to run on multiple systems. But python3 doesn't work on my Windows, use python?

rffontenelle: That won't be necessary, as [actions/setup-python](https://github.com/actions/setup-python) installs python3 binary (or symlink to binary, I don't know), so those commands work fine in the CI environment. See [this Windows workflow run in my fork](https://github.com/rffontenelle/python-docs-theme/actions/runs/10221587570/job/28284437047).

Any documentation we might write will probably need to take your scenario in consideration.

Wulian233: Great🚀

AA-Turner, line 74: ``python`` is more typical in GHA (& consistent with other jobs in this file). I don't think setuptools is needed?
```suggestion
          pip install -U pip
          pip install Babel jinja2
      - run: python babel_runner.py extract
      - run: python babel_runner.py init -l pt_BR
      - run: python babel_runner.py update
      - run: python babel_runner.py update -l pt_BR
      - run: python babel_runner.py compile
      - run: python babel_runner.py compile -l pt_BR
```

babel_runner.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/venv python3
+"""Script for handling translations with Babel"""
+from __future__ import annotations
+
+import argparse
+import os
+import subprocess
+
+import tomllib
```
AA-Turner, line 9:
```suggestion
import argparse
import os
import subprocess
import tomllib
```

babel_runner.py
```diff
@@ -0,0 +1,114 @@
...
+
+# Global variables used by pybabel below
+DOMAIN = "messages"
+COPYRIGHT_HOLDER = "Python Software Foundation"
+LOCALES_DIR = os.path.relpath(os.path.join(PROJECT_DIR, "locales"))
+POT_FILE = os.path.relpath(os.path.join(LOCALES_DIR, f"{DOMAIN}.pot"), PROJECT_DIR)
+SOURCE_DIR = os.path.relpath(
+    os.path.join(PROJECT_DIR, "python_docs_theme"), PROJECT_DIR
+)
+MAPPING_FILE = os.path.relpath(os.path.join(PROJECT_DIR, ".babel.cfg"), PROJECT_DIR)
```
AA-Turner, line 21: It might be nicer to use `pathlib` here? Also, any reason to use relative paths rather than absolute?

.github/workflows/tests.yml
rffontenelle: Installing babel and jinja2 without setuptools made extraction fail with `Unknown extraction method 'jinja2' babel`. I even reproduced locally and only solved after a pip's --force-reinstall

AA-Turner: Oh, I remember -- `babel` uses setuptools' `Command` class internally. Hence the mess that is ``babel_runner.py`` in Sphinx...

A

babel_runner.py
rffontenelle: I added this blank line because of ruff linter: https://github.com/python/python-docs-theme/actions/runs/10208999238/job/28250436250

AA-Turner: Try setting python-version for Ruff -- it is likely defaulting to something before 3.11

rffontenelle: You are correct. Ruff uses python-version from pyproject.toml, which is currently set to ">= 3.8".

Should I set ruff's target-version permanently, patch it for the workflow environment, or is there another way of passing the Python version to ruff via CLI?

AA-Turner: https://docs.astral.sh/ruff/configuration/#the-config-cli-flag

``--config "target-version=py311"``

.github/workflows/tests.yml
hugovk: We already use `python` rather than `python3` in _this_ repo, so I suggest we keep it the same here.

rffontenelle: Ok, change applied. It looks like there are 3 votes on `python` (AA-Turner too) :D

babel_runner.py
rffontenelle: I was using relative path because the output would be too long because of the absolute path. By using relative paths I achieved `creating catalog locales/pt_BR/LC_MESSAGES/messages.po based on locales/messages.pot` (just 'locales/...' instead of the absolute path).

But using the feedback from both of you, I replaced os with pathlib and used `cwd` keyword argument of `subprocess.run()` in a way that this whole relative path pollution is no longer necessary.

babel_runner.py
rffontenelle: Well, the problem is that ruff check is failing is not just on my end, but in lint.yml action run, so AFAIK I'm not able to set the python version unless is replace python-version or set target-version in pyproject.toml.

I've worked around this issue using tomli as alternative (used your suggestion in babel add TOML support pull request). For this I added requirements.txt to use as single point of installing tomli or not based on Python version.

hugovk (commented):
A few nits, then I think we can merge this and iterate if needed, unless anyone else has any last comments?

Wulian233:
In the Simplified Chinese documentation, there is an issue regarding hard-coded strings such as "added in version" and "deprecated in version" etc. that cannot be translated. Can this PR solve this problem?

https://github.com/python/python-docs-zh-cn/issues/364

rffontenelle:
> In the Simplified Chinese documentation, there is an issue regarding hard-coded strings such as "added in version" and "deprecated in version" etc. that cannot be translated. Can this PR solve this problem?
> 
> https://github.com/python/python-docs-zh-cn/issues/364

No, that's a Sphinx string and it is not translated to zh_CN: https://github.com/sphinx-doc/sphinx/blob/116a430caca98580c8877cf7b0ba270eae9604bf/sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po#L1951

Consider contributing in https://explore.transifex.com/sphinx-doc/sphinx-1/

babel_runner.py
```diff
@@ -0,0 +1,119 @@
...
+        print(f"There is already a message catalog for locale {locale}, skipping.")
+        return
+    cmd = ["pybabel", "init", "-i", POT_FILE, "-d", LOCALES_DIR, "-l", locale]
+    subprocess.run(cmd, cwd=PROJECT_DIR, check=True)
+
+
+def update_catalogs(locale: str):
+    """Update translations from existing message catalogs"""
+    cmd = ["pybabel", "update", "-i", POT_FILE, "-d", LOCALES_DIR]
+    if locale != "":
```
ezio-melotti, line 77:
```suggestion
    if locale:
```

babel_runner.py
```diff
@@ -0,0 +1,119 @@
...
+        cmd.extend(["-l", locale])
+    subprocess.run(cmd, cwd=PROJECT_DIR, check=True)
+
+
+def compile_catalogs(locale: str):
+    """Compile existing message catalogs"""
+    cmd = ["pybabel", "compile", "-d", LOCALES_DIR]
+    if locale != "":
```
ezio-melotti, line 85:
```suggestion
    if locale:
```

babel_runner.py
```diff
@@ -0,0 +1,119 @@
...
+        help="language code (needed for init, optional for update and compile)",
+    )
+
+    args = parser.parse_args()
+    locale = args.locale if args.locale else ""
+
+    if args.command == "extract":
+        extract_messages()
+    elif args.command == "init":
+        if locale == "":
```
ezio-melotti, line 109:
```suggestion
        if not locale:
```

babel_runner.py
```diff
@@ -0,0 +1,119 @@
...
+        help="command to be executed",
+    )
+    parser.add_argument(
+        "-l",
+        "--locale",
+        help="language code (needed for init, optional for update and compile)",
+    )
+
+    args = parser.parse_args()
+    locale = args.locale if args.locale else ""
```
ezio-melotti, line 104:
```suggestion
    parser.add_argument(
        "-l",
        "--locale",
        default="",
        help="language code (needed for init, optional for update and compile)",
    )

    args = parser.parse_args()
    locale = args.locale
```
This should be equivalent.

hugovk (approved):
Thank you!

Wulian233:
It looks great! I have a question: where should we translate this theme, and how will the translation automatically sync to this repository?

rffontenelle:
@Wulian233 I mentioned that in the issue and it is yet to be defined.

Some projects coordinate translation efforts in the own forge via pull/merge requests — not optimal for the translators. There is also translation platforms like Weblate, Transifex and others.

My vote would be to use Weblate, which provides "libre hosting" for free software and features automatic opening and updating pull request for contributed translations. Tasks involve: setting up Weblate project (I can help with/do it), maintainers here update POT and give time before releasing to receive translations; and merging translation pull requests from Weblate.

PR #198, opened 2024-09-27
Generate digital attestations for PyPI (PEP 740)

hugovk:
PEP 740 ("Index support for digital attestations") introduces signatures which links the PyPI package to the GitHub repo, and helps users verify the source and authenticity of packages.

PyPI is still implementing support, but we can already start using it, which should also help them test out.

* https://peps.python.org/pep-0740/
* https://github.com/pypa/gh-action-pypi-publish#generating-and-uploading-attestations
* https://github.com/pypi/warehouse/issues/15871

### DIFF
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -11,6 +11,9 @@ on:
 permissions:
   contents: read
 
+env:
+  FORCE_COLOR: 1
+
 jobs:
   # Always build & lint package.
   build-package:
@@ -45,3 +48,5 @@ jobs:
 
       - name: Upload package to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          attestations: true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 build-backend = "flit_core.buildapi"
 requires = [
-  "flit_core>=3.7",
+  "flit-core>=3.7",
 ]
 
 [project]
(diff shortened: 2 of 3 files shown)


.github/workflows/pypi-package.yml
```diff
@@ -45,3 +48,5 @@ jobs:
 
       - name: Upload package to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          attestations: true
```
ezio-melotti, line 52: So this is what actually generates the attestations, and even though the feature is still experimental we can still enable it here (and in the other repos) and let `gh-action-pypi-publish` deal with potential API changes, right?

I guess the other changes are unrelated, but it was convenient to bundle them together.

pyproject.toml
ezio-melotti, line 1: Just out of curiosity, do you know what caused the changes below?
Some of them seem backward, at least coming from Python (e.g. removing the double space before `#` and adding spaces within `[]` and `{}`), at least coming from Python.

ezio-melotti (approved):

pyproject.toml
hugovk: They nearly all came from the pyproject-fmt upgrade from 1.x to 2.x.

We needed to bump actionlint so it wouldn't fail on the new attestations config, then I updated the other pre-commit at the same time.

One pyproject-fmt change is to align all the comments. Before, we just so happened to manually align them. Now, it automatically does it. I'm not sure why pyproject-fmt chose a single space, but it's a TOML file, not a Python file.

(Looks like it's using this [TOML formatter](https://taplo.tamasfe.dev/configuration/formatter-options.html) which has `align_comments` but no option for how many spaces.)

---

I manually adjusted this into a single line:
```toml
include = [
    "python_docs_theme/",
]
```
to:
```toml
include = [ "python_docs_theme/" ]
```

---

The extra spaces are also from pyproject-fmt, and common in examples in the spec: https://toml.io/en/v1.0.0

Again, this TOML not Python. I guess I might slightly prefer more a Python-y style, but having an autoformatter keep things consistent, and easier to compare across projects, is much more valuable for me.

ezio-melotti: Yep, I figured that TOML might be following different conventions, and it makes sense to be consistent with those.
Thanks for looking into this!

Issue #203, opened 2024-10-31
Search box cover up text in PDFs


gabaryza:
# Documentation

Python 3.13.0

I attempted to save a documentation web page, the beginning of the Language Reference, as a PDF for reading offline. In the second and subsequent pages of the saved PDF, the search box from the top of the web page overlays the first line(s) of each PDF page rendering them unreadable.

Demo: https://python-docs-theme-previews--204.org.readthedocs.build/en/204/

hugovk:
Thanks!

It looks like this for page 1:

<img width="1102" alt="image" src="https://github.com/user-attachments/assets/c6f1708e-8ce9-4609-9551-d0445d8075df">


And is worse for subsequent pages:

<img width="1106" alt="image" src="https://github.com/user-attachments/assets/ea7766db-0551-4b6a-b195-5b1b20715c99">

hugovk:
Please see https://github.com/python/python-docs-theme/pull/204 for a fix, and https://python-docs-theme-previews--204.org.readthedocs.build/en/204/ for a demo.

gabaryza:
Thank you.

PR #204, opened 2024-10-31
Hide header and search bar when printing

hugovk:
Fixes https://github.com/python/python-docs-theme/issues/203.

Using media query for print, hide the header that contains the search bar.


hugovk:
Screenshot of PDF with fix, page 1:

image

Subsequent page:

image

Wulian233 (approved):

Issue #209, opened 2024-12-20
More responsive/wider sidebar


tomasr8:
The current sidebar is quite narrow and on [some pages](https://docs.python.org/3.14/library/importlib.html) cuts off quite a lot of the text:

![image](https://github.com/user-attachments/assets/2ddcdd8a-e1f6-43f7-b7b3-67469f847b5b)

Some possible solutions:

- Increase the sidebar width (and maybe center the main content?)
- Make the sidebar resizable

ONKARBH:
@tomasr8 

"Hi, I’m Onkar Bhosale, and I’m interested in contributing to the improvement of the sidebar’s responsiveness and width. While I’m still learning Python, I have significant experience in frontend technologies such as HTML, CSS, and JavaScript. I believe I can contribute to this issue by adjusting the sidebar’s width and making it more responsive across various devices. I am eager to collaborate on this project and enhance the user experience of the documentation theme. Looking forward to the opportunity to contribute!"

Wulian233:
Welcome to contribute! You can fork this repository, make your commit and then create a pull request for the maintainers to review

Onkar9890:
Hi @tomasr8,

I’d like to take up this issue and work on enhancing the sidebar’s responsiveness and width. With my experience in frontend development using HTML, CSS, and JavaScript, I am confident in implementing the suggested solutions and testing them across various devices.

Could you please assign this issue to me? I’ll ensure timely updates and collaboration to achieve the desired improvements.

Thank you!

Wulian233:
> I’d like to take up this issue and work on enhancing the sidebar’s responsiveness and width. With my experience in frontend development using HTML, CSS, and JavaScript, I am confident in implementing the suggested solutions and testing them across various devices.
> 
> Could you please assign this issue to me? I’ll ensure timely updates and collaboration to achieve the desired improvements.
> 
> Thank you!

It's still the same as what was mentioned above😂😂😂😂

> I’m interested in contributing to the improvement of the sidebar’s responsiveness and width. While I’m still learning Python, I have significant experience in frontend technologies such as HTML, CSS, and JavaScript. I believe I can contribute to this issue by adjusting the sidebar’s width and making it more responsive across various devices. I am eager to collaborate on this project and enhance the user experience of the documentation theme. Looking forward to the opportunity to contribute!

terryjreedy:
[...]

PR #214, opened 2025-01-25
Make the sidebar resizable

tomasr8:
Addresses #209 

- Sidebar is no longer collapsible but is instead resizable with a slider
- Bumps the width to `250px` (happy to bikeshed the width)
- Uses flexbox for the sidebar content and sidebar button instead of floats
- Adds some ARIA-specific attributes to the slider

Feedback welcome!

Preview: https://python-docs-theme-previews--214.org.readthedocs.build/en/214/

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -203,7 +203,7 @@ div.sphinxsidebar input[type='text'] {
     margin-left: 0;
     color: #444444;
     font-size: 1.2em;
-    cursor: col-resize;
+    cursor: default; /* cursor is set to 'col-resize' using JS */
     padding-top: 1px;
     float: none;
     display: table;
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -139,7 +139,13 @@ span.pre {
     white-space: unset;
 }
 
+div.bodywrapper {
+    margin-left: 250px;
+}
+
 div.sphinxsidebar {
+    display: flex;
+    width: 250px;
     float: none;
     position: sticky;
     top: 0;
@@ -157,12 +163,12 @@ div.sphinxsidebar h4 {
 }
 
 div.sphinxsidebarwrapper {
-    width: 217px;
     box-sizing: border-box;
     height: 100%;
     overflow-x: hidden;
     overflow-y: auto;
-    float: left;
+    float: none;
+    flex-grow: 1;
 }
 
 div.sphinxsidebarwrapper > h3:first-child {
@@ -197,12 +203,13 @@ div.sphinxsidebar input[type='text'] {
     margin-left: 0;
     color: #444444;
     font-size: 1.2em;
-    cursor: pointer;
+    cursor: col-resize;
     padding-top: 1px;
-    float: right;
+    float: none;
     display: table;
     /* after Sphinx 4.x and earlier is dropped, only the below is needed */
     width: 12px;
+    min-width: 12px;
     border-radius: 0 5px 5px 0;
     border-left: none;
 }
@@ -468,7 +475,7 @@ div.genindex-jumpbox a {
         font-size: 0.875rem;
     }
     div.bodywrapper {
-        margin: 0;
+        margin: 0 !important;
     }
     /* Typography */
     div.body h1 {
@@ -490,7 +497,7 @@ div.genindex-jumpbox a {
     }
     /* Remove sidebar and top related bar */
     div.related, .sphinxsidebar {
-        display: none;
+        display: none !important;
     }
     /* Anchorlinks are not hidden by fixed-positioned navbar when scrolled to */
     html {
--- a/python_docs_theme/static/sidebar.js_t
+++ b/python_docs_theme/static/sidebar.js_t
@@ -49,6 +49,9 @@ const initialiseSidebar = () => {
   const sidebarButton = document.createElement("div")
   sidebarButton.id = "sidebarbutton"
 {% endif %}
+  const sidebarMinWidth = 200
+  const sidebarMaxWidth = Math.round(0.5 * window.innerWidth)
+
   sidebarbutton.innerHTML = ""
   sidebarbutton.tabindex = "0"  // make it focusable
   sidebarbutton.role = "slider"
@@ -60,7 +63,10 @@ const initialiseSidebar = () => {
   function onMouseMove(e) {
     e.preventDefault()
     const sidebarWidth = sidebar.offsetWidth
-    const newWidth = Math.max(0, Math.min(window.innerWidth, sidebarWidth + e.clientX - clientX))
+    const newWidth = Math.max(
+      sidebarMinWidth,
+      Math.min(sidebarMaxWidth, sidebarWidth + e.clientX - clientX)
+    )
     clientX = e.clientX
     sidebar.style.width = `${newWidth}px`
     bodyWrapper.style.marginLeft = `${newWidth}px`
--- a/python_docs_theme/static/sidebar.js_t
+++ b/python_docs_theme/static/sidebar.js_t
@@ -26,6 +26,7 @@
  */
 
 const initialiseSidebar = () => {
+  const ngettext = Documentation.ngettext
   // global elements used by the functions.
   const bodyWrapper = document.querySelector(".bodywrapper")
   const sidebar = document.querySelector(".sphinxsidebar")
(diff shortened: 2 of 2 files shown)


python_docs_theme/static/sidebar.js_t
```diff
@@ -44,46 +44,43 @@ const initialiseSidebar = () => {
...
-  sidebarArrow.innerText =  "«"
-  sidebarButton.appendChild(sidebarArrow)
-  sidebar.appendChild(sidebarButton)
 {% endif %}
+  sidebarbutton.innerHTML = ""
+  sidebarbutton.tabindex = "0"  // make it focusable
+  sidebarbutton.role = "slider"
+  sidebarbutton.title = _("Resize sidebar")
+  sidebarbutton.setAttribute("aria-label", _("Resize sidebar by dragging"))
+  sidebarbutton.setAttribute("aria-valuetext", _("Sidebar width XXX pixels"))
```
tomasr8, line 57: I didn't manage to find `ngettext`, is it available?

hugovk: `_` is an alias for `Documentation.gettext` in Sphinx's `doctools.js`:

https://github.com/mne-tools/mne-tools.github.io/blob/1c89caf568b03d2571239053ad05b1a084f6252a/dev/_static/doctools.js#L146-L147

Which also contains `Documentation.ngettext`, you could try that?

hugovk (commented):
I wonder if we should have a min and/or max width, or let the user have a tiny or massive sidebar if they really want?

I will note that PR this does make it impossible to completely collapse the sidebar, which is possible by clicking the divider at https://docs.python.org/3/. Is that something we want to keep?

tomasr8:
> I wonder if we should have a min and/or max width

I agree, it might be nicer to have some limits. I'll add some so we can try it out

> I will note that PR this does make it impossible to completely collapse the sidebar, which is possible by clicking the divider at https://docs.python.org/3/. Is that something we want to keep?

I removed it because I felt it wouldn't really have a much of a purpose anymore, though if people prefer both we could make it work :)

tomasr8:
I added a min/max width for the sidebar (200px minimum and a half of `window.innerWidth` maximum but we can refine the values)

python_docs_theme/static/pydoctheme.css
```diff
@@ -197,12 +203,13 @@ div.sphinxsidebar input[type='text'] {
     margin-left: 0;
     color: #444444;
     font-size: 1.2em;
-    cursor: pointer;
+    cursor: col-resize;
```
hugovk, line 206: I tested with JavaScript disabled, and the sidebar was static, as expected, but the cursor made it look resizable.

Should we use the default cursor (not `pointer`), and switch to `col-resize` using JS?

So then it only changes when it's possible to resize.

tomasr8: I didn't consider the case when JS is disabled, but if that's a supported use case, changing the cursor using JS sounds like the way to go :+1:

tomasr8: Added :) Would you mind having another look?

hugovk: Cursor looks good 👍

hugovk:
Left: with JS
Right: without JS, still showing the "«" symbol, making it look collapsible, but it's not. Can we remove it?

image

tomasr8:
> Right: without JS, still showing the "«" symbol, making it look collapsible, but it's not. Can we remove it?

Definitely, I need to get into the mindset that JS might not be available :sweat_smile: Currently I'm removing it through JS, but a CSS rule should work even better. I'll fix it later today

ezio-melotti:
I'm not sure making the sidebar resizable is the right solution to the problem described in #209.

The problem I was solving when I made the sidebar collapsible many years ago, was to improve the experience on small screen (mostly phones) where the sidebar was taking up over half of the screen.  Making it collapsible is also useful on regular screens when the user doesn't need it and wants more space for the actual documentation.

This PR takes away the collapsibility and replaces it with resizeability which is IMO much less useful.

Going back to the original problems listed in the issue:
1. Content doesn't always fit
   * I think a better solution for this would be to use CSS ellipsis possibly combined with a `title=""` or something similar to show the full text on hover (even though last time I checked there wasn't an easy way to achieve this with CSS only).  I'd also argue that this is functionally a non-issue and at best an aesthetic one, since the visible part is most often enough to recognize what you are looking for (assuming you are not already using `ctrl+f`).  Changing `overflow` and/or `word-wrap` rules, increasing sidebar `width`, decreasing `text-indent`, reducing `font-size` are also viable solutions.
2. Sidebar is too narrow on wide screens
   * This could be solved either with a `%`-based width, or with media queries

I'd would be ok with a resizable sidebar if it didn't affect collapsibility, but offering both features might just end up complicating the UI for little gain IMO.

hugovk:
> The problem I was solving when I made the sidebar collapsible many years ago was to improve the experience on small screen (mostly phones), where the sidebar was taking up over half of the screen.

At least since the theme was made responsive (in https://github.com/python/python-docs-theme/pull/46) the sidebar is hidden on screens narrower than 1024 pixels, so this part shouldn't be a problem any more.

> Going back to the original problems listed in the issue:

I'm more favourable to making the sidebar wider than trying to display more stuff in a small space. Screens are getting wider. 

Truncating text has usability/accessibility issues (can't ctrl+f), as does showing text on hover (can't hover on touch devices, there's more [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role#accessibility_concerns)).

python_docs_theme/static/sidebar.js_t
```diff
@@ -44,46 +45,64 @@ const initialiseSidebar = () => {
...
-    sidebarArrow.innerText = "»"
-    sidebarButton.title = _("Expand sidebar")
-    window.localStorage.setItem("sidebar", "collapsed")
-  }
+  sidebarbutton.innerHTML = ""
+  sidebarbutton.tabindex = "0"  // make it focusable
+  sidebarbutton.role = "slider"
+  sidebarbutton.title = _("Resize sidebar")
+  sidebarbutton.style.cursor = "col-resize" // Set the cursor only if JS is enabled
+  sidebarbutton.setAttribute("aria-label", _("Resize sidebar by dragging"))
```
trallard, line 61: Noting that `aria-label` should only be used like so to provide additional usage context to assistive technology users.
After testing this it seems the resizing function is only operable via dragging (mouse or similar) so it seems this `aria-label` is used as a "catch all" to provide usage instructions, thus defeating the purpose of ARIA, it might be best removing it altogether, and instead add a comment.

python_docs_theme/static/sidebar.js_t
```diff
@@ -44,46 +45,64 @@ const initialiseSidebar = () => {
...
-  const collapse_sidebar = () => {
-    bodyWrapper.style.marginLeft = ".8em"
-    sidebar.style.width = ".8em"
-    sidebarWrapper.style.display = "none"
-    sidebarArrow.innerText = "»"
-    sidebarButton.title = _("Expand sidebar")
-    window.localStorage.setItem("sidebar", "collapsed")
-  }
+  sidebarbutton.innerHTML = ""
+  sidebarbutton.tabindex = "0"  // make it focusable
```
trallard, line 57: Noting that I could not focus this while navigating with a keyboard nor through screen reader so there might be something tripping this.

trallard:
I agree with @ezio-melotti  here, I am not convinced this approach addresses the issues from the original issue. I think a much simpler alternative would be to make the sidebar wider (with a minimum width or similar) while keeping the collapsing function as this is a feature folks used when multitasking or in smaller screens.

The slider/splitter added in this PR adds a lot more complexity, and is only usable with a mouse so adds a new layer of accessibility concerns. There would be significant changes needed to make it keyboard and assistive tech operable and we might need keeping the collapsing feature anyway leading to a complex UI which might prove tricky to get right.

(Note that the collapse button as is needs some rework to make it operable with multiple inputs, but we can work on that separately but would take a lot less work than making an accessible slider + accessible collapse button).

tomasr8:
> Noting that I could not focus this while navigating with a keyboard nor through screen reader so there might be something tripping this.

Strange, it's working fine for me locally. Maybe some issue with the preview?

Anyway, thanks for your input! I see your points and I agree that adding a slider might be too disruptive to some workflows. Personally, I'd also be happy with just increasing the (max-)width of the sidebar so that it's a bit more flexible.

> [...] this is a feature folks used when multitasking or in smaller screens.

Out of curiosity, what kind of set up are we talking about? The sidebar is hidden for widths <1024 pixels and for screens larger than that there's enough space for the text even with the sidebar.

In any case, if nobody disagrees, I can close this PR and open one that simply makes the sidebar wider :)

PR #216, opened 2025-01-27
Note minimum requirements for Sphinx

AA-Turner:
We implicitly support the [versions used in CPython](https://github.com/python/docsbuild-scripts?tab=readme-ov-file#check-current-version). However, Python 3.9 uses a pinned version of this theme (https://github.com/python/cpython/pull/115351), so we can explicitly document our minimum supported versions as those used in the Python 3.10 branch.

A

----
📚 Documentation preview 📚: https://python-docs-theme-previews--216.org.readthedocs.build/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Python Docs Sphinx Theme
 
 This is the theme for the Python documentation.
+It requires Python 3.9 or newer and Sphinx 3.4 or newer.
 
 Note that when adopting this theme, you're also borrowing an element of the
 trust and credibility established by the CPython core developers over the
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,17 +1,17 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.8
+    rev: v0.9.3
     hooks:
       - id: ruff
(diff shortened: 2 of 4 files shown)


hugovk (approved):

README.md
```diff
@@ -1,6 +1,7 @@
 # Python Docs Sphinx Theme
 
 This is the theme for the Python documentation.
+It requires Python 3.9 or newer and Sphinx 3.4 or newer.
```
hugovk, line 4: In another project we noticed the Python version mentioned in docs said 3.6+, but in reality it's 3.9+ and we'd forgotten to update that three times!

To DRY, perhaps we don't need to document this in words, seeing as we have definitive metadata in `pyproject.toml`? 

That is:

```toml
requires-python = ">=3.9"
...
dependencies = [
  "sphinx>=3.4",
]
```

AA-Turner: I suppose I put it in the README to make it very clear -- perhaps we could replace it with a policy statement instead? I.e.:

"It supports all Sphinx versions required by the [stable](https://devguide.python.org/versions/) Python branches."

I'm not a huge fan of this wording, though -- it feels clunky.

hugovk: Well, it's okay to leave this in the README if you like, at least it's the repo's first page and not some other "hidden"/forgotten docs page.

PR #218, opened 2025-02-02
Make sidebar width more flexible

tomasr8:
Related issue: https://github.com/python/python-docs-theme/issues/209

This is a less disruptive version of https://github.com/python/python-docs-theme/pull/214
This simply increases the width of the sidebar to 260px (not set in stone, happy to discuss the appropriate value)

Two more things I considered that we could discuss but didn't include in the PR
- Adding another breakpoint beyond 1024px which would further increase the sidebar width (not sure if static breakpoints are the way to go though)
- Making the sidebar more flexible by only defining min/max-width and letting the contents determine the exact value. This would require a larger change to the CSS as the sidebar is currently static (it relies on a fixed margin on `.bodywrapper`). Though we could convert it to flexbox as well.

Feedback welcome!

----
📚 Documentation preview 📚: https://python-docs-theme-previews--218.org.readthedocs.build/

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -140,6 +140,8 @@ span.pre {
 }
 
 div.sphinxsidebar {
+    display: flex;
+    width: 260px;
     float: none;
     position: sticky;
     top: 0;
@@ -156,13 +158,17 @@ div.sphinxsidebar h4 {
     margin-top: 1.5em;
 }
 
+div.documentwrapper > div.bodywrapper {
+    margin-left: 260px;
+}
+
 div.sphinxsidebarwrapper {
-    width: 217px;
     box-sizing: border-box;
     height: 100%;
     overflow-x: hidden;
     overflow-y: auto;
-    float: left;
+    float: none;
+    flex-grow: 1;
 }
 
 div.sphinxsidebarwrapper > h3:first-child {
@@ -199,10 +205,11 @@ div.sphinxsidebar input[type='text'] {
     font-size: 1.2em;
     cursor: pointer;
     padding-top: 1px;
-    float: right;
+    float: none;
     display: table;
     /* after Sphinx 4.x and earlier is dropped, only the below is needed */
     width: 12px;
+    min-width: 12px;
     border-radius: 0 5px 5px 0;
     border-left: none;
 }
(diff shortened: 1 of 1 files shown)


AA-Turner:
Could we do a percentage width of say 25% (so 256px at the lower bound before the sidebar disappears)?

tomasr8:
Just to clarify, you mean to have a fixed sidebar width of 256px or some kind of dynamic sizing? The latter would require additional style changes

AA-Turner:
A width defined in % or vw units, yes.

tomasr8:
We might need some more breakpoints for that - `25vw` might be too much on very wide screens

hugovk:
And maybe an upper bound where it's a fixed width, compare:

https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/index.html

---

As it stands, this is PR good for mobile, the sidebar is always shown:

<details>
<summary>Screenshot</summary>


python-docs-theme-previews--218 org readthedocs build_en_218_(iPhone SE)



</details>

tomasr8:
I made the sidebar more like the pydata theme Hugo shared. The width is now calculated as `min(25vw, 350px)` so it is at maximum `350px` and at minimum `256px` (just before it gets hidden). 

> As it stands, this is PR good for mobile, the sidebar is always shown:

Fixed, some of the selectors I was using were too specific.

python_docs_theme/static/pydoctheme.css
```diff
@@ -140,6 +140,8 @@ span.pre {
 }
 
 div.sphinxsidebar {
+    display: flex;
```
AA-Turner, line 143: Is flexbox required here?

tomasr8: This allows me to remove the hardcoded `width: 217px;` from `div.sphinxsidebarwrapper`

hugovk:
(There's a conflict after merging https://github.com/python/python-docs-theme/pull/219.)

tomasr8:
> (There's a conflict after merging #219.)

Will fix it in a bit

AA-Turner (approved):

hugovk:
Thanks!

tomasr8:
Thanks for the reviews/comments!

PR #219, opened 2025-02-02
Horizontally align the collapse button icon

tomasr8:
Just something that was bugging me a bit :D I hope this won't be an issue regarding 4.x compatibility?

Before
image

After
image

----
📚 Documentation preview 📚: https://python-docs-theme-previews--219.org.readthedocs.build/

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -191,6 +191,9 @@ div.sphinxsidebar input[type='text'] {
 }
 
 #sidebarbutton {
+    display: flex;
+    justify-content: center;
+    align-items: center;
     /* Sphinx 4.x and earlier compat */
     height: 100%;
     background-color: #CCCCCC;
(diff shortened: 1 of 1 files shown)


m-aciek:
It may be worth checking how this affect Sphinx 4-, as Python 3.10 and 3.9 use Sphinx 3.4.3.

python_docs_theme/static/pydoctheme.css
```diff
@@ -200,19 +203,12 @@ div.sphinxsidebar input[type='text'] {
...
     padding-top: 1px;
     float: right;
-    display: table;
     /* after Sphinx 4.x and earlier is dropped, only the below is needed */
     width: 12px;
     border-radius: 0 5px 5px 0;
     border-left: none;
 }
 
-#sidebarbutton span {
```
Wulian233, line 210: Can we achieve compatibility without deleting this?

AA-Turner:
> It may be worth checking how this affect Sphinx 4-, as Python 3.10 and 3.9 use Sphinx 3.4.3.

Note I'm proposing pinning `python-docs-theme` for the 3.10 and 3.11 branches in https://github.com/python/cpython/pull/129576. We already pinned for the 3.8 and 3.9 branches in https://github.com/python/cpython/issues/115349. However in the absence of that, this PR should be tested with 3.4.3, 7.2.6, and 8.1.3 ([[ref]](https://github.com/python/docsbuild-scripts/?tab=readme-ov-file#check-current-version)).

A

tomasr8:
I'll test it later today or tomorrow (at Fosdem currently)

tomasr8:
In addition to 8.1.3, I checked with 3.4.3 and 7.2.6 and both work as well :)

hugovk (approved):

PR #220, opened 2025-02-02
Change root url to docs.python.org

### LINKED ISSUE #217: Docs Python.org button
# Documentation

Below 1024 pixels width there is no button that returns you to the docs home.

Image

____

For screen widths > 1023 you can click the "3.13.1 Documentation »" button to return to the docs index. 

Image

# Proposal

When using the docs on a small screen there is no easy way to return to the index as you can see in the below page:

Image

I propose for usability the Python logo returns you to the index, this is generally the web standard.

StanFromIreland:
Closes #217 

----
📚 Documentation preview 📚: https://python-docs-theme-previews--220.org.readthedocs.build/


python_docs_theme/theme.conf
```diff
@@ -30,7 +30,7 @@ hosted_on =
 issues_url =
 license_url =
 root_name = Python
-root_url = https://www.python.org/
+root_url = https://www.docs.python.org/
```
AA-Turner, line 33: This changes the default for every user of this theme. Should we use [`html_theme_options`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_theme_options)?

StanFromIreland: We could, also the link has to be docs.python.org as for some reason it is otherwise broken.

hugovk:
## Mobile

image

The logo always links to `docs.python.org`. That's okay when we're looking the 3.13 docs, but if we're looking at, say, 3.11 or 3.14, we'll end up at a different docs set which would be confusing.

Can it point to the same docs homepage we're looking at?

## Desktop

With this change, these circled links both point to the docs homepage:

image

Before, the first one pointed to `www.python.org`, and the chevrons "»" indicated a hierarchy. Is it a bit odd that the upper level link is the same as the next one?

Or I guess, the top one is `docs.python.org` and the next is `docs.python.org/3.x`, so that's kind of a hierarchy, for the default case of  `docs.python.org` >> `docs.python.org/3`, but maybe not really for, say, `docs.python.org` >>  `docs.python.org/3.11`?

And do we want to lose the link to `www.python.org`?

The logo doesn't point to anything here.

StanFromIreland:
I don't see a need for a button to python.org, it is linked when necessary.

python_docs_theme/theme.conf
StanFromIreland: Also coming back to this option, what do we set the default as?

hugovk:
We discussed this in the monthly docs workgroup meeting and decided to replace the `www.python.org` link with a `docs.python.org` one. 

We can add a link under "Project information:" to `www.python.org`.

---

Not brought up in the meeting, but as mentioned in https://github.com/python/python-docs-theme/pull/220#issuecomment-2629560831, let's make sure the link doesn't go to the default docs.python.org for all docs versions; it should go to the homepage of that version.

For example, if on this page, the link should go to that page:

* https://docs.python.org/3/library/hashlib.html -> https://docs.python.org/3/
* https://docs.python.org/3.11/library/hashlib.html -> https://docs.python.org/3.11/
* https://docs.python.org/dev/library/hashlib.html -> https://docs.python.org/dev/

m-aciek:
Closing. We need something more robust per Hugo's comment:

> let's make sure the link doesn't go to the default docs.python.org for all docs versions

Issue #154, opened 2023-08-30
Proposal to add copy button to code examples [copy of gh-107024]


churnikov:
Hi!

I'm planning to move [this work](https://github.com/python/cpython/issues/107024) to this repo.

@AA-Turner Suggested to me in this PR https://github.com/python/cpython/pull/107037#issuecomment-1654256388 to do that. Please tell me, what do you think :)
Do you agree that that's a good idea?

If so, I guess I would need to rewrite this function, right?
https://github.com/python/python-docs-theme/blob/main/python_docs_theme/static/copybutton.js

AA-Turner:
E.g. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard; https://clipboardjs.com/ -- there are many ways to and libraries for interacting with the clipboard -- ought we abandon our JavaScript, ought we adopt a pre-existing library, ought we use the browser APIs, etc.

A

ferdnyc:
FYI, there's also a pre-made [`sphinx_copybutton`](https://github.com/executablebooks/sphinx-copybutton) extension available, that uses `clipboard.js`.

tomasr8:
If there's still interest in having this feature, I can give it a go. I would suggest to use [`navigator.clipboard`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) which has been supported by all major browsers for >5 years at this point.

We also already have a function which can strip the prompt markers from the code so the implementation (bar the UI) should be relatively simple.

ferdnyc:
@tomasr8 That function also removes the output lines from interactive sessions, which I think is an important part of being able to copy/paste commands from those sessions.

If I might be permitted a bit of feature creep, I notice that the prompt characters in the listings are styled `user-select: none` so that they don't get caught up in manual copy actions.

Rather than having the "strip out the prompts and output" button, it might be easier to just style the _output_ as non-selectable as well, so that even manual selects would only get the input lines. (The copy button could then copy only the same elements, automatically.)

It's not too useful if selecting and copying this:
```python
>>> 2 + 2
4
>>> 50 - 5*6
20
>>> (50 - 5*6) / 4
5.0
>>> 8 / 5  # division always returns a floating-point number
1.6
```

gets you _this_ on the clipboard:
```python
2 + 2
4
50 - 5*6
20
(50 - 5*6) / 4
5.0
8 / 5  # division always returns a floating-point number
1.6
```

[...]

PR #231, opened 2025-04-12
Add a copy button to code samples

tomasr8:
I kept the style similar to the previous button. For now it's text but maybe we could use an icon instead?
Note that unlike the previous button, I added the copy button to all code samples, not only the interactive ones.

I'd like to know what you think!

Before:
image

After:
image
image

Single line:
image
image

[copybtn.webm](https://github.com/user-attachments/assets/b955f7b2-17e0-4b07-8ed8-74380450c4fd)

----
📚 Documentation preview 📚: https://python-docs-theme-previews--231.org.readthedocs.build/

### DIFF
--- a/python_docs_theme/static/copybutton.js
+++ b/python_docs_theme/static/copybutton.js
@@ -34,14 +34,14 @@ const loadCopyButton = () => {
         let timeout = null
         // define the behavior of the button when it's clicked
         return event => {
-          clearTimeout(timeout)
-          const buttonEl = event.currentTarget
-          const codeEl = buttonEl.nextElementSibling
-          navigator.clipboard.writeText(getCopyableText(codeEl))
-          buttonEl.innerText = _("Copied!")
-          timeout = setTimeout(() => {
-              buttonEl.innerText = _("Copy")
-          }, 1500)
+            clearTimeout(timeout)
+            const buttonEl = event.currentTarget
+            const codeEl = buttonEl.nextElementSibling
+            navigator.clipboard.writeText(getCopyableText(codeEl))
+            buttonEl.innerText = _("Copied!")
+            timeout = setTimeout(() => {
+                buttonEl.innerText = _("Copy")
+            }, 1500)
         }
     }
 
--- a/python_docs_theme/static/copybutton.js
+++ b/python_docs_theme/static/copybutton.js
@@ -33,11 +33,23 @@ const loadCopyButton = () => {
     const makeOnButtonClick = () => {
         let timeout = null
         // define the behavior of the button when it's clicked
-        return event => {
+        return async event => {
+            // check if the clipboard is available
+            if (!navigator.clipboard || !navigator.clipboard.writeText) {
+                return;
+            }
+
             clearTimeout(timeout)
             const buttonEl = event.currentTarget
             const codeEl = buttonEl.nextElementSibling
-            navigator.clipboard.writeText(getCopyableText(codeEl))
+
+            try {
+                await navigator.clipboard.writeText(getCopyableText(codeEl))
+            } catch (e) {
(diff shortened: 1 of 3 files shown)


tomasr8:
The docsbuild CI seems broken, could it be related to this change: https://github.com/python/docsbuild-scripts/commit/2113fd7daaed924663c986384048defcab68b342 ?

hugovk:
Yep, fixed here in https://github.com/python/python-docs-theme/pull/232. I've updated the branch from `main`.

python_docs_theme/static/copybutton.js
```diff
@@ -1,65 +1,47 @@
...
-            for (const el of getHideableCopyButtonElements(codeEl)) {
-                el.hidden = false
-            }
-            buttonEl.title = hide_text
-            buttonEl.dataset.hidden = "false"
+        return event => {
+          clearTimeout(timeout)
+          const buttonEl = event.currentTarget
+          const codeEl = buttonEl.nextElementSibling
+          navigator.clipboard.writeText(getCopyableText(codeEl))
```
hugovk, line 40: My IDE gives a warning that the returned promise is ignored:

image

Should we have a `.then()` and only do "Copied!" in there?

https://web.dev/patterns/clipboard/copy-text

And maybe also log any error, in case it happens?

hugovk (commented):
Thanks for the PR!

It's working well on a quick test on Chrome, Safari, Firefox and Opera on macOS, and Chrome on Android.

> For now it's text but maybe we could use an icon instead?

An advantage of text over icons is it's more accessible and less open to misinterpretation and different cultural associations.

However, GitHub and Discourse use a double-square icon, so this may be familiar:

```python
a = 1
b = 2
assert a < b
```

image

https://discuss.python.org/t/pep-781-make-type-checking-a-built-in-constant/85728/1 

And when clicked, GitHub shows a tick and "Copied", Discourse shows "copied!". They also use aria-labels for accessibility.

Discord is similar and shows a tick when clicked:

image

But I'm fine with this text-only version, and an icon could also be a followup if we want that.


https://github.com/python/python-docs-theme/issues/154#issuecomment-2780671338:

> I would suggest to use [`navigator.clipboard`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) which has been supported by all major browsers for >5 years at this point.

Yes, that feels long enough, it has [~95% support](https://caniuse.com/mdn-api_navigator_clipboard) and we don't need to support copy for older browsers. But what happens on a browser without support? As long as we don't break things completely.

python_docs_theme/static/copybutton.js
tomasr8: Thanks for testing it!

> They also use aria-labels for accessibility.

I think that won't be necessary in our case? We're using a regular button with `type="button"` and text content.

> But what happens on a browser without support?

It's gonna raise inside the click handler which shouldn't break anything else but I'll add some proper feature detection.

> Should we have a .then() and only do "Copied!" in there?

Yep, that sounds like a safer option!

tomasr8:
Looks like there's an issue with the docs workflow:
```
ERROR: Cannot install python-docs-theme==2025.2 and sphinx==3.4.3 because these package versions have conflicting dependencies.
```

hugovk:
Also happening on `main`:

https://github.com/python/python-docs-theme/actions/runs/14452159103/job/40527148242#step:6:32

Not a new failure, but it only started failing the build recently! Here's the previous `main` build with an error but the CI is green 😅 

https://github.com/python/python-docs-theme/actions/runs/14428653230/job/40460802134#step:6:32

Anyway, don't worry about it for this PR :)

hugovk (approved):
Thanks!

PR #236, opened 2025-04-25
Add support for Python 3.14

hugovk:
Also test building the CPython branches on the corresponding versions.

----
📚 Documentation preview 📚: https://python-docs-theme-previews--236.org.readthedocs.build/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: 3
+          python-version: ${{ matrix.branch == 'origin/main' && '3.14' || matrix.branch }}
           allow-prereleases: true
           cache: pip
       - name: Clone docsbuild scripts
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: 3
+          python-version: ${{ matrix.branch == 'origin/main' && '3.14' || matrix.branch }}
```
AA-Turner, line 20: I think we can safely use `3.14` in the matrix? I'd prefer to avoid complex logic like this if possible.

AA-Turner (approved):

.github/workflows/tests.yml
hugovk: Updated: https://github.com/python/python-docs-theme/pull/236/commits/4d59c56e668341304a68c8098a6c4caba7f5a393

There's now a build error because it's failing to build Pillow from source, because it has no wheels for 3.14 yet:

https://github.com/python/python-docs-theme/actions/runs/14679936811/job/41201341521?pr=236

Which actually indicates the previous commit was not actually building: it took 0 seconds:

https://github.com/python/python-docs-theme/actions/runs/14667652291/job/41166360664?pr=236

think Pillow is being installed because it's a dependency of matplotlib. What's a good way to skip matplotlib for 3.14?

PR #237, opened 2025-04-25
Replace deprecated classifier with licence expression (PEP 639)

hugovk:
See https://hugovk.dev/blog/2025/improving-licence-metadata/

----
📚 Documentation preview 📚: https://python-docs-theme-previews--237.org.readthedocs.build/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,14 +8,14 @@ requires = [
 name = "python-docs-theme"
 description = "The Sphinx theme for the CPython docs and related projects"
 readme = "README.md"
-license.file = "LICENSE"
+license = "PSF-2.0"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -8,14 +8,14 @@ requires = [
 name = "python-docs-theme"
 description = "The Sphinx theme for the CPython docs and related projects"
 readme = "README.md"
-license.file = "LICENSE"
+license = "PSF-2.0"
+license-files = [ "LICENSE" ]
```
AA-Turner, line 12: I don't think we need to set this, flit has a sensible default. Doesn't hurt, though.

AA-Turner (approved):

PR #240, opened 2025-04-29
Fix copy button with multiple tracebacks

tomasr8:
The copy button currently removes everything after a traceback, including code that should be copied: https://docs.python.org/3.14/whatsnew/3.14.html#improved-error-messages


tomasr8:
Seems to work now: https://python-docs-theme-previews--240.org.readthedocs.build/en/240/whatsnew/3.14.html#improved-error-messages

hugovk:
Thanks!

PR #243, opened 2025-05-21
Make the copy button only appear when hovered over or clicked on

kevteg:
- Made the copy button only appear when hovered over or clicked on
- Tested on chrome, arc, safari and mobile devices

----
📚 Documentation preview 📚: https://python-docs-theme-previews--243.org.readthedocs.build/en/243/library/string.html#module-string


hugovk (approved):
Thanks, works well in Chrome on macOS and Android.

PR #246, opened 2025-08-07
Enable theme translation

StanFromIreland:
Adds Polish to start. Based on @rffontenelle 's pr. This does not allow for translation of the Copy button for some reason, but I think that can be left for another time.

Result:

image

----
📚 Documentation preview 📚: https://python-docs-theme-previews--246.org.readthedocs.build/

[🇵🇱 Polish preview 🇵🇱](https://stan.ulbrych.org/preview/theme-translation/)


rffontenelle:
Thanks for fixing what was missing with the solution!

m-aciek (commented):
Translation improvement suggestions, otherwise looks good to me 👍

python_docs_theme/__init__.py
```diff
@@ -12,9 +13,22 @@
 THEME_PATH = Path(__file__).resolve().parent
 
 
+def setup_translations(app):
+    translation = gettext.translation(
+        domain="messages",
+        localedir=str(THEME_PATH / "locales"),
```
m-aciek, line 19:
```suggestion
        localedir=os.fspath(THEME_PATH / "locales"),
```

m-aciek (commented):
Suggestion for better portability

babel_runner.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/venv python3
+#!/usr/bin/env python3
```
hugovk, line 1: We're running this via `python babel_runner.py` in CI, shall we remove the shebang and executable bit?

```suggestion
```

babel_runner.py
```diff
@@ -15,9 +15,9 @@
 # Global variables used by pybabel below (paths relative to PROJECT_DIR)
 DOMAIN = "messages"
 COPYRIGHT_HOLDER = "Python Software Foundation"
-LOCALES_DIR = "locales"
-POT_FILE = Path(LOCALES_DIR, f"{DOMAIN}.pot")
 SOURCE_DIR = "python_docs_theme"
```
hugovk, line 18: Is the `"python_docs_theme"` on line 13 also an instance of `SOURCE_DIR`? Should it be:
```
INIT_PY = PROJECT_DIR / SOURCE_DIR / "__init__.py"
```

hugovk (commented):
Looks about right, but I'd appreciate a PR approval from a translator as well.

And we can always iterate and fix as needed after an initial release.

python_docs_theme/__init__.py
```diff
@@ -12,9 +14,22 @@
 THEME_PATH = Path(__file__).resolve().parent
 
 
+def setup_translations(app):
+    translation = gettext.translation(
```
AA-Turner, line 18: Should we use [`sphinx.locale.get_translation()`](https://www.sphinx-doc.org/en/master/extdev/i18n.html#sphinx.locale.get_translation) here? 

https://www.sphinx-doc.org/en/master/extdev/i18n.html#extension-internationalization-i18n-and-localization-l10n-using-i18n-api

python_docs_theme/locales/pl_PL/LC_MESSAGES/messages.mo
AA-Turner, line 1: Do we need to commit these binary files? My understanding is that Sphinx will create them from the `.po` files during the build process.

hugovk: Good point.

To compare another project, https://github.com/python-humanize/humanize only has `.po` files committed:

https://github.com/python-humanize/humanize/tree/main/src/humanize/locale

And the binary `.mo` files are generated during the release (by [msgfmt](https://github.com/python-humanize/humanize/blob/0e027051034b7dcb4eb65aded290cf56a89d8aa9/.github/workflows/release.yml#L37) and not Sphinx, but perhaps that's an implementation detail).

StanFromIreland:
> Looks about right, but I'd appreciate a PR approval from a translator as well.

Maciek has already reviewed translation, see above.

python_docs_theme/locales/pl_PL/LC_MESSAGES/messages.mo
StanFromIreland: It wasn't built by default, when I was testing locally.

python_docs_theme/locales/pl_PL/LC_MESSAGES/messages.po
hugovk, line 1: Let's change the directory name to `/locale/` singular.

That looks much more common: Sphinx has singular; 234 vs 29 in 7.5k top repos; 35 vs 0 in a bunch of repos I have cloned.

StanFromIreland: Sure, though the Sphinx docs actually use `locales`

python_docs_theme/locales/pl_PL/LC_MESSAGES/messages.mo
rffontenelle: > It wasn't built by default, when I was testing locally.

It is not. Humanize runs a script in the release CI workflow to generate. See [this commit](https://github.com/python-humanize/humanize/blob/0e027051034b7dcb4eb65aded290cf56a89d8aa9/.github/workflows/release.yml#L37) and the [called script](https://github.com/python-humanize/humanize/blob/main/scripts/generate-translation-binaries.sh).

StanFromIreland: I can open a separate PR for the release script.

StanFromIreland:
[Preview!](https://stan.ulbrych.org/preview/theme-translation/) (built with `-D language='pl'`)

StanFromIreland:
It is no longer a follow up, it is now included in this PR!

Verified manually that the [artefact](https://github.com/python/python-docs-theme/actions/runs/17437990048?pr=246) has the compiled files.

rffontenelle:
Sweet! Was setup-python really necessary, considering that python is already included in the GH runner image?

hugovk (approved):
Thanks!

PR #258, opened 2025-09-08
Remove 'navigation'

methane:
I found this in Transifex. This is not displayed. (e.g. `display: none`).

----
📚 Documentation preview 📚: https://python-docs-theme-previews--258.org.readthedocs.build/

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -3,7 +3,6 @@
 {%- macro relbar() %}
 {# modified from sphinx/themes/basic/layout.html #}
     <div class="related" role="navigation" aria-label="related navigation">
-      <h3>{{ _('Navigation') }}</h3>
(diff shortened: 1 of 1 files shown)


encukou:
It might be there for screen readers.

methane:
Generally, screen readers ignores `display: none`.
`div.related h3 { display: none }` lives since first commit of Sphinx.
https://github.com/sphinx-doc/sphinx/commit/d60ca8d49d1654968886612883a2affa9dccfaa8#diff-9b474a3f571175e217231911be3927a1e66e79b84b4a98c95a4788b475c604bdR68

methane:
Perhaps they're using h3 tags to indicate that this section is the place for navigation links when the css isn't loaded.
If so, it would be best to translate this part as well, just like the displayed text.

PR #262, opened 2025-09-09
Add German translation

cmaureir:


StanFromIreland (approved):
We don’t have a German translation of the documentation, so this won’t be used, but it’s good to have it IMO. LGTM

PR #264, opened 2025-09-09
Add Korean translation

StanFromIreland:
[Pulled from TX](https://app.transifex.com/python-doc/python-docs-theme/language/ko/)


AA-Turner:
Can we use the TX integration instead of having to make arbitrary PRs?

StanFromIreland:
Yes, that's what I said in the issue? I can't enable it however, and we need to generate a TX config file and place it in the directory.

StanFromIreland:
I've also added the freshly translated Indonesian!

PR #267, opened 2025-09-14
Remove unused image

StanFromIreland:
This file was added in the initial commit, however only `py.svg` is used now.


StanFromIreland:
I don't think this needs to be included in the CHANGELOG, right?

PR #269, opened 2025-09-14
Remove trailing space from link

StanFromIreland:
Current:
image

With PR:
image

----
📚 Documentation preview 📚: https://python-docs-theme-previews--269.org.readthedocs.build/

### DIFF
--- a/python_docs_theme/layout.html
+++ b/python_docs_theme/layout.html
@@ -137,12 +137,12 @@
 {% block footer %}
     <div class="footer">
     &copy; {% if theme_copyright_url or hasdoc('copyright') -%}
-        <a href="{% if theme_copyright_url %}{{ theme_copyright_url }}{% else %}{{ pathto('copyright') }}{% endif %}">
-      {%- endif -%}
-      {%- trans -%}Copyright{%- endtrans -%}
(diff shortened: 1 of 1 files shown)


StanFromIreland:
That's interesting, I can also remove the leading one.

python_docs_theme/layout.html
```diff
@@ -137,12 +137,9 @@ <h3>{{ _('Navigation') }}</h3>
 {% block footer %}
     <div class="footer">
     &copy; {% if theme_copyright_url or hasdoc('copyright') %}
-      <a href="{% if theme_copyright_url %}{{ theme_copyright_url }}{% else %}{{ pathto('copyright') }}{% endif %}">
-    {% endif %}
-    {% trans %}Copyright{% endtrans %}
-    {% if theme_copyright_url or hasdoc('copyright') %}
-      </a>
-    {% endif %} {{ copyright|e }}.
+      {# Wrapping the below link introduces spurious whitespace #}
```
hugovk, line 140: We can keep the wrapped version and add minus symbols to tell it to delete whitespace from that side:

```
    {%- endif %}
    {%- trans %}BCopyright{% endtrans %}{% if theme_copyright_url or hasdoc('copyright') %}</a>
```

PR #271, opened 2025-09-16
Update authors in `pyproject.toml`

StanFromIreland:
----
📚 Documentation preview 📚: https://python-docs-theme-previews--271.org.readthedocs.build/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ description = "The Sphinx theme for the CPython docs and related projects"
 readme = "README.md"
 license = "PSF-2.0"
 license-files = [ "LICENSE" ]
-authors = [ { name = "python-docs-theme maintainers" } ]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -10,7 +10,7 @@ description = "The Sphinx theme for the CPython docs and related projects"
 readme = "README.md"
 license = "PSF-2.0"
 license-files = [ "LICENSE" ]
-authors = [ { name = "PyPA", email = "distutils-sig@python.org" } ]
+authors = [ { name = "python-docs-theme maintainers" } ]
```
hugovk, line 13:
```suggestion
authors = [ { name = "Python Docs Sphinx Theme maintainers" } ]
```

hugovk (commented):
> The lint rule exceptions are unnecessary since there are no infractions.

No infractions *right now*. If we remove the exceptions, a new PR can fail and the author will have to change their code to meet these rules. We don't want that.

PR #282, opened 2025-11-26
Add support for Python 3.15

hugovk:
The earlier approach was something like:
```yml
      matrix:
        branch: ["main", "3.14", "3.13", "3.12"]
...
          python-version: ${{ matrix.branch == 'main' && '3.15' || matrix.branch }}
...
          --branches ${{ matrix.branch }}
```
Which do you prefer?

----
📚 Documentation preview 📚: https://python-docs-theme-previews--282.org.readthedocs.build/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,12 +12,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        branch: ["3.14", "3.13", "3.12"]
+        python-version: ["3.15", "3.14", "3.13", "3.12"]
+        include:
+          - python-version: "3.15"
+            branch: "main"
     steps:
       - uses: actions/checkout@v5
       - uses: actions/setup-python@v6
         with:
-          python-version: ${{ matrix.branch }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -12,12 +12,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        branch: ["3.14", "3.13", "3.12"]
+        python-version: ["3.15", "3.14", "3.13", "3.12"]
+        include:
+          - python-version: "3.15"
+            branch: "main"
```
ezio-melotti: A possible improvement is to compute the branch once and assign it to an env var:
```suggestion
        include:
          - python-version: "3.15"
            branch: "main"
      env:
        # set the branch name to either 'main' or the version number
        BRANCH: ${{ matrix.branch || matrix.version }}
```
This keeps the logic closer and avoids duplication, but since we only need the branch once in the workflow, it might be overkill.

ezio-melotti (approved):

PR #296, opened 2026-02-13
Add styling rule to field-list for vertical alignment

### LINKED ISSUE #289: Fix vertical alignment of definition terms
The vertical alignment of "Editors" inside its box is a bit off, it's too high up: https://docs.python.org/3/whatsnew/3.14.html

Image

This is a `<dl><dt>Editors</dt><dd>...</dd>` structure.

We should also check other uses of this HTML in the docs.

WangRemy:
Fixes #289 

# Before
Before

# After
After

----
📚 Documentation preview 📚: https://python-docs-theme-previews--296.org.readthedocs.build/

### DIFF
--- a/python_docs_theme/static/pydoctheme_dark.css
+++ b/python_docs_theme/static/pydoctheme_dark.css
@@ -77,6 +77,7 @@ span.highlighted {
 /* Below for most things in text */
 
 dl.field-list > dt {
+    align-self: center;
     background-color: #434;
 }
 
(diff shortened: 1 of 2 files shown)


python_docs_theme/static/pydoctheme_dark.css
```diff
@@ -77,6 +77,7 @@ span.highlighted {
 /* Below for most things in text */
 
 dl.field-list > dt {
+    align-self: center;
```
StanFromIreland: Note that this is redundant, it is inherited anyway. But, as Hugo noted, this is currently broken either way.

StanFromIreland:
Closed in favour of #303.

PR #305, opened 2026-04-17
Add styling for `soft-deprecated` class

StanFromIreland:
For the new directive proposed in https://github.com/python/cpython/pull/148630. I have to use two separate colours to meet accessibility standards.

In light mode with "Poison Purple Paradise" (per [some site](https://colornamer.robertcooper.me/)), we have a ratio of 4.92:

image

In dark mode with "Jacaranda Pink", we have a ratio of 5.22:

image

----
📚 Documentation preview 📚: https://python-docs-theme-previews--305.org.readthedocs.build/en/305/library/re.html#re.match

### DIFF
--- a/python_docs_theme/static/pydoctheme.css
+++ b/python_docs_theme/static/pydoctheme.css
@@ -789,10 +789,12 @@ div.genindex-jumpbox a {
     --versionadded: var(--good-color);
     --versionchanged: var(--middle-color);
     --deprecated: var(--bad-color);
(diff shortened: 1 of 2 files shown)


hugovk:
Will review once https://github.com/python/cpython/pull/148630 is merged so it's easier to test.

hugovk (approved):

StanFromIreland:
Thanks for the review!

PR #318, opened 2026-07-24
Remove old zizmor config file

hugovk:
No longer need to ignore any rules after removing `documentation-links.yml` in https://github.com/python/python-docs-theme/pull/308.


StanFromIreland (approved):
LGTM

PR #307, opened 2026-04-24
Hash pin GitHub Actions

hugovk:
Yet another compromise via unpinned GitHub Actions: https://socket.dev/blog/bitwarden-cli-compromised

Let's hash-pin GHA for our PyPI packages.

Done via `uvx gha-update`.

----
📚 Documentation preview 📚: https://python-docs-theme-previews--307.org.readthedocs.build/

PR #319, opened 2026-07-27
Prepare 2026.7 release

hugovk:
It's been three months since the last release and we have a couple of things waiting.


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [2026.7](https://github.com/python/python-docs-theme/releases/tag/2026.7)
+
+* Add Vietnamese translation by @StanFromIreland in https://github.com/python/python-docs-theme/pull/309
+* Switch to Hatchling build hooks for i18n packaging and exclude .po from wheels by @StanFromIreland in https://github.com/python/python-docs-theme/pull/312
```
StanFromIreland, line 6: Let's also add the fix I just merged :-)

hugovk: Not sure if https://github.com/python/python-docs-theme/pull/320 is that interesting to end users of the theme?

I also left out:

* Hash pin GitHub Actions by @hugovk in https://github.com/python/python-docs-theme/pull/307
* Replace deprecated action with RtD app by @hugovk in https://github.com/python/python-docs-theme/pull/308
* Remove old zizmor config file by @hugovk in https://github.com/python/python-docs-theme/pull/318

StanFromIreland: I think the Makefile is a little more interesting than updating GitHub Actions, people who build from source might be interested in such changes. But, if you think it's not significant enough, LGTM without it as well.

PR #323, opened 2026-08-16
Bump `hynek/build-and-inspect-python-package` and `pypa/gh-action-pypi-publish` for Metadata 2.5

StanFromIreland:
See the discussion in https://github.com/pypa/hatch/issues/2292 for more information. Hatching produces Metadata 2.5, but twine (bundled by these Actions) needs to be updated to handle it. This causes [CI failures](https://github.com/python/python-docs-theme/actions/runs/31947325656/job/95165531677?pr=322) like:

```
  Checking /tmp/baipp/dist/python_docs_theme-2026.7-py3-none-any.whl: ERROR    InvalidDistribution: Invalid distribution metadata: '2.5' is not a     
           valid metadata version                                                 
```


hugovk (approved):
Thanks!

PR #324, opened 2026-08-17
Add RTL support

sepehr-rs:
Added RTL support to python-docs-theme. Passing the `is_rtl=true` flag when building the documentation now generates RTL-friendly pages.
I also added Vazirmatn as the font for RTL scripts (it really improves readability), but I’m not sure how acceptable this is, so I’m happy to change it if the maintainers disagree.
Please let me know if any changes or additions are needed. Thanks in advance!


StanFromIreland:
I think it's better to do the font in a follow up, since I have some concerns about licensing and such things.

sepehr-rs:
> I think it's better to do the font in a follow up, since I have some concerns about licensing and such things.

Although I’m okay with leaving this for a follow-up, [Vazirmatn](https://github.com/rastikerdar/vazirmatn) is an open-source project, licensed under the SIL Open Font License. But if you still think this is better handled in a follow-up, I’ll do that. Thanks a lot!

StanFromIreland:
> But if you still think this is better handled in a follow-up

I'd prefer that, personally.

StanFromIreland:
Can you please update your branch to pull in the translation?

python_docs_theme/static/pydoctheme.css
```diff
@@ -251,19 +284,10 @@ div.body {
 }
 
 div.code-block-caption {
-    background-color: #eee;
```
StanFromIreland, line 254: Why do we loose these?

sepehr-rs: I think this was removed mistakenly. I will bring it back now. Thanks for catching it!

StanFromIreland:
Some observations:

image

The hamburger on the mobile view is on the wrong side.

image

Function signatures are still LTR, but on the right side?

image

Inline function references are LTR, but have the parentheses on the RTL side.

image

Similar to previous point.

image

On English pages, the text is also now LTR. Is this alright from your perspective? I think Sphinx now has some tags to differentiate translated messages, so we could avoid this if necessary.

image

There's now a white artefact beside the sidebar?

StanFromIreland:
Also, I tried to get Claude to look at this. Here's its analysis:

> Breaks RTL outright
> 
> 1. Mobile nav double-reverses — pydoctheme.css:632, :694
> In a direction: rtl container a plain flex-direction: row already lays out right-to-left. Adding row-reverse flips it back, and .toggler__label { order: 1 } drops the hamburger at the far left. Net result: the mobile header renders in exactly the LTR arrangement the PR set out to mirror. Delete both rules.
> 
> 2. Search icon vs. its gutter — :663, :679
> Line 657 correctly converts padding-left: 24px → padding-inline-start. Lines 662–664 then set inline-start: 4px; inline-end: 24px, whose used values are identical to the original physical rule. But the magnifier svg is position: absolute with no inset, so it falls at its static position — main-start, i.e. the right edge in RTL. Icon right, 24px gutter left, placeholder text under the icon.
> 
> 3. 230px phantom gutter on RTL desktop — :179
> div.bodywrapper { margin-inline-start: 0 } maps to margin-right in RTL, so classic's margin: 0 0 0 230px (classic.css_t:41, sidebarwidth=230) survives untouched. Every RTL desktop page ≥1024px gets dead space left and the sidebar right. The removed margin-left: min(25vw,350px) used to cover both. Use margin: 0.
> 
> Unmirrored / ineffective
> 
> 4. direction: ltr is a no-op on inline code — :266
> Per CSS, direction applies to an inline box only when unicode-bidi isn't normal. It works for the block-level pre/.highlight, but inline literals like os.path.join(a, b) still get bidi-reordered. Needs unicode-bidi: isolate alongside.
> 
> 5. API signatures missed entirely — :265
> Sphinx emits signatures as <dt class="sig sig-object py"> with <span class="sig-paren"> — no pre, no code, no .highlight. So os.path.join(path, *paths) renders with mirrored parens and reordered params. .sig, dl > dt, .sig-param need the same treatment.
> 
> 6. RTL rule out-specifies the mobile override — :241
> html[dir='rtl'] div.body is (0,2,2); the @media (max-width:1023px) { div.body { padding: 0 } } at :566 is (0,1,1) and media queries add no specificity, so RTL wins regardless of order — 1.2em of padding on a min-width: 100% element, i.e. overflow. LTR is unaffected, so it's an RTL-only asymmetry. padding-inline-start on the base rule avoids the extra selector.
> 
> 7. Leftovers: .menu { padding: … 20px } (:758) unconverted; div.sphinxsidebar ul ul { margin-left: 20px } (basic.css) never overridden, so RTL toctree indents away from parents; dl.field-list > dt padding unmirrored.
> 
> 8. Inline style beats every stylesheet — layout.html:9
> style="margin-right: 10px" on the first li.right. The PR floats these left in RTL, so the gap lands between items instead of at the edge. → margin-inline-end.
> 
> Design-level
> 
> 9. RTL is JS-only — layout.html:85
> Mutating documentElement.dir client-side means JS-disabled → fully LTR, a hard failure rather than degraded styling. This theme already has an html-page-context handler (__init__.py:22) that can set html_tag server-side, and derive RTL from config.language (ar, he, fa, ur) instead of a second option that can silently disagree with it.
> 
> 10. theme_is_rtl|lower == "true" — layout.html:83
> Reimplements Sphinx's tobool filter, more narrowly. {"is_rtl": "1"} or "yes" — both idiomatic — silently do nothing. Use |tobool.
> 
> 11. Float-based breadcrumbs — :128
> div.related li { display: inline } already mirrors for free under direction: rtl. Floating computes display: block, killing the collapsible whitespace (hence the compensating padding-inline, which then also hits li.right), and breaks .switchers' inline-flex.
> 
> 12. transition: inset-inline-start 400ms — :744
> An unrecognised transition-property invalidates the whole shorthand, so the menu snaps instead of sliding — and it's not RTL-gated, so LTR builds are exposed too. Animating an inset also forces layout per frame; transform: translateX() composites.
> 
> Unrelated regression
> 
> 13. div.code-block-caption lost all seven declarations — :286
> Not an RTL change. With the override gone, classic's { color: #efefef; background-color: #1c4e63 } wins — captions become dark-blue bars instead of the theme's grey box, no border, no radius, no padding. The surviving & + div pre { border-top-*-radius: 0 !important } then squares off a code block under a caption with no matching box, and dark mode's border-color: #616161 is inert with no border-style declared. Looks accidental.
> 
> 14. Dead code — div.document { flex-direction: row } (:146) is the initial value; .mobile-nav { inset-inline-start: 0 } (:610) is inert. Four /* was: … */ comments annotate 4 of ~25 conversions at random, and one replaced /* follows div.body pre */, which documented a real cross-rule colour dependency.

sepehr-rs:
Thanks so much for the thorough review! I’ll work through these comments and send a patch soon.

sepehr-rs:
Hi @StanFromIreland, I addressed the review comments you left with Claude's help, built the docs locally to verify the changes, and all the issues you pointed out appear to be resolved. I also checked that the LTR version remains unaffected.
Please let me know if anything else needs adjustment. Thanks again for the thorough review!

StanFromIreland:
> I also checked that the LTR version remains unaffected.

Indeed, testing with the current patch, the only change is the addition of `dir="ltr"` attribute:

<details>

<summary>A selection of pages</summary>

```diff
diff --git a/html/whatsnew/changelog.html b/base/html/whatsnew/changelog.html
index 903b246..c60dcd2 100644
--- a/html/whatsnew/changelog.html
+++ b/base/html/whatsnew/changelog.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html lang="en" data-content_root="../">
+<html lang="en" dir="ltr" data-content_root="../">
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" /><meta name="viewport" content="width=device-width, initial-scale=1" />
@@ -20,7 +20,7 @@
     
     <link rel="stylesheet" type="text/css" href="../_static/pygments.css?v=b86133f3" />
     <link rel="stylesheet" type="text/css" href="../_static/classic.css?v=234b1a7c" />
-    <link rel="stylesheet" type="text/css" href="../_static/pydoctheme.css?v=4365c8fe" />
+    <link rel="stylesheet" type="text/css" href="../_static/pydoctheme.css?v=83ab0a67" />
     <link id="pygments_dark_css" media="(prefers-color-scheme: dark)" rel="stylesheet" type="text/css" href="../_static/pygments_dark.css?v=0fc419ee" />
     
     <script src="../_static/documentation_options.js?v=d86cccaa"></script>
@@ -50462,7 +50462,7 @@ Kilburn.</p></li>
 <a href="https://www.python.org/psf/donations/">Please donate.</a>
 <br>
     <br>
-      Last updated on Aug 19, 2026 (20:32 UTC).
+      Last updated on Aug 19, 2026 (20:30 UTC).
     
       <a href="/bugs.html">Found a bug</a>?
```

```diff
diff --git a/html/library/sqlite3.html b/base/html/library/sqlite3.html
index 6decdde..fc9f358 100644
--- a/html/library/sqlite3.html
+++ b/base/html/library/sqlite3.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html lang="en" data-content_root="../">
+<html lang="en" dir="ltr" data-content_root="../">
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" /><meta name="viewport" content="width=device-width, initial-scale=1" />
@@ -20,7 +20,7 @@
     
     <link rel="stylesheet" type="text/css" href="../_static/pygments.css?v=b86133f3" />
     <link rel="stylesheet" type="text/css" href="../_static/classic.css?v=234b1a7c" />
-    <link rel="stylesheet" type="text/css" href="../_static/pydoctheme.css?v=4365c8fe" />
+    <link rel="stylesheet" type="text/css" href="../_static/pydoctheme.css?v=83ab0a67" />
     <link id="pygments_dark_css" media="(prefers-color-scheme: dark)" rel="stylesheet" type="text/css" href="../_static/pygments_dark.css?v=0fc419ee" />
     
     <script src="../_static/documentation_options.js?v=d86cccaa"></script>
@@ -2979,7 +2979,7 @@ statements.  This is no longer the case.</p>
 <a href="https://www.python.org/psf/donations/">Please donate.</a>
 <br>
     <br>
-      Last updated on Aug 19, 2026 (20:32 UTC).
+      Last updated on Aug 19, 2026 (20:30 UTC).
     
       <a href="/bugs.html">Found a bug</a>?
```

```diff
diff --git a/html/library/datetime.html b/base/html/library/datetime.html
index b32b4a9..0381fa5 100644
--- a/html/library/datetime.html
+++ b/base/html/library/datetime.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html lang="en" data-content_root="../">
+<html lang="en" dir="ltr" data-content_root="../">
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" /><meta name="viewport" content="width=device-width, initial-scale=1" />
@@ -20,7 +20,7 @@
     
     <link rel="stylesheet" type="text/css" href="../_static/pygments.css?v=b86133f3" />
     <link rel="stylesheet" type="text/css" href="../_static/classic.css?v=234b1a7c" />
-    <link rel="stylesheet" type="text/css" href="../_static/pydoctheme.css?v=4365c8fe" />
+    <link rel="stylesheet" type="text/css" href="../_static/pydoctheme.css?v=83ab0a67" />
     <link id="pygments_dark_css" media="(prefers-color-scheme: dark)" rel="stylesheet" type="text/css" href="../_static/pygments_dark.css?v=0fc419ee" />
     
     <script src="../_static/documentation_options.js?v=d86cccaa"></script>
@@ -3649,7 +3649,7 @@ for a good explanation.</p>
 <a href="https://www.python.org/psf/donations/">Please donate.</a>
 <br>
     <br>
-      Last updated on Aug 19, 2026 (20:32 UTC).
+      Last updated on Aug 19, 2026 (20:30 UTC).
     
       <a href="/bugs.html">Found a bug</a>?
```

</details>

StanFromIreland:
@cben, maybe you can help review this? From what I gather you're the expert in this area, I'd appreciate your thoughts here.

python_docs_theme/__init__.py
```diff
@@ -37,12 +44,37 @@ def combined(message: str) -> str:
...
 
 
+def add_html_dir_to_context(
+    app: Sphinx,
+    pagename: str,
+    templatename: str,
+    context: dict[str, Any],
+    doctree: None,
+) -> None:
+    language = app.config.language or "en"
```
StanFromIreland: Why not `context["language"]`?

python_docs_theme/__init__.py
```diff
@@ -18,6 +19,12 @@
 MESSAGE_CATALOG_NAME = "python-docs-theme"
 
 
+def _tobool(val: object) -> bool:
```
StanFromIreland: This seems quite overcomplicated for something used once.

python_docs_theme/__init__.py
```diff
@@ -37,12 +44,37 @@ def combined(message: str) -> str:
...
+    if is_rtl_option in (None, ""):
+        is_rtl = False
+    else:
+        is_rtl = _tobool(is_rtl_option)
+
+    dir_attr = "rtl" if is_rtl else "ltr"
+
+    content_root = context.get("content_root", "")
+    lang_part = f' lang="{escape(language)}"' if language is not None else ""
+    context["html_tag"] = Markup(
```
StanFromIreland: I'm not a fan of this, what if Sphinx adds an attribute?

python_docs_theme/static/pydoctheme.css
```diff
@@ -211,11 +238,15 @@ div.sphinxsidebar input[type='text'] {
 }
 
 div.body {
-    padding: 0 0 0 1.2em;
+    padding-inline-start: 1.2em;
```
StanFromIreland, line 240: This no longer zeros the other sides:

https://github.com/sphinx-doc/sphinx/blob/e44a40eb2f810558ccd9da1425421270ccb81351/sphinx/themes/classic/static/classic.css.jinja#L40

python_docs_theme/static/pydoctheme.css
```diff
@@ -258,12 +303,10 @@ div.code-block-caption {
     padding: 0.3em 0.6em;
     font-size: 90%;
     color: #333;
-
```
StanFromIreland, line 261: These are unrelated changes.

python_docs_theme/static/pydoctheme.css
```diff
@@ -477,17 +520,12 @@ div.footer a:hover {
 }
 
 
-dl > dt span ~ em,
-.sig {
-    font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
```
StanFromIreland, line 482: Why are we dropping the font?

python_docs_theme/static/pydoctheme.css
```diff
@@ -416,9 +459,9 @@ table.footnote td {
 
 div.footer {
     line-height: 150%;
-    text-align: right;
+    text-align: end; /* was: text-align: right */
```
StanFromIreland:
```suggestion
    text-align: end;
```
These comments don't add any useful information.

python_docs_theme/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from pathlib import Path
 
+from markupsafe import Markup, escape
```
StanFromIreland: This is now an undeclared direct dependency...

StanFromIreland (commented):
I'm a little worried about regressing LTR, we should be very careful here.

Would something in `layout.html` like:

```html
{% block body_tag %}<body{% if theme_is_rtl|tobool %} dir="rtl"{% endif %}>{% endblock %}
```
not work instead of `add_html_dir_to_context`?

sepehr-rs:
Thanks a lot for the review! I addressed your review comments, but for the whitespace changes, they don't show up anymore on my code editor, so I hope they're resolved. Please let me know if any changes are needed :)

cben:
@StanFromIreland Ooh I'll be glad to review but have family events & vacations for a week, not sure worth waiting for me but I can review post-merge too.

python_docs_theme/static/pydoctheme.css
```diff
@@ -117,6 +117,17 @@ div.related ul::after {
     display: block;
 }
 
+div.related ul {
+    padding: 0;
+    padding-inline-start: 10px;
+}
+
+html[dir='rtl'] div.related ul li.right {
```
StanFromIreland: The `dir` attribute isn't set on `html` so these won't work.

StanFromIreland:
I've created https://github.com/python/python-docs-theme/pull/325 to get RtD previews with an RTL build.

StanFromIreland:
> @StanFromIreland Ooh I'll be glad to review but have family events & vacations for a week, not sure worth waiting for me but I can review post-merge too.

That would be great!

python_docs_theme/static/pydoctheme.css
StanFromIreland: I still see these as removed.

python_docs_theme/static/pydoctheme.css
```diff
@@ -695,29 +737,36 @@ div.genindex-jumpbox a {
...
     }
 
     /* Sliding side menu */
     .menu-wrapper {
         display: block;
         position: fixed;
         top: 0;
-        transition: left 400ms ease;
-        left: -310px;
+        inset-inline-start: 0;
```
StanFromIreland, line 749: The menu appears on screen even when closed, and on the wrong side:

Image

python_docs_theme/static/pydoctheme.css
sepehr-rs: Okay, I tried fixing these by adding spaces in place of the removed whitespace, but I don't know if this has fixed the problem or not. Could this be a quirk on GitHub's side?

python_docs_theme/static/pydoctheme.css
sepehr-rs: Thanks for catching this. Changing `html[dir='rtl']` to `body[dir='rtl']` fixed this.

StanFromIreland:
Oh sorry, I meant to push to https://github.com/python/python-docs-theme/pull/325.

python_docs_theme/static/pydoctheme.css
```diff
@@ -627,22 +672,22 @@ div.genindex-jumpbox a {
...
         height: 100%;
         box-shadow: none;
         border: 0;
-        border-left: 1px solid #a9a9a9;
+        border-inline-start: 1px solid #a9a9a9;
         cursor: pointer;
-        margin-right: 0;
+        margin-inline-end: 0;
     }
     .nav-content .search svg {
```
StanFromIreland, line 684: This switch to an absolute position breaks LTR:

Image

python_docs_theme/static/pydoctheme.css
```diff
@@ -695,29 +740,36 @@ div.genindex-jumpbox a {
...
     /* Sliding side menu */
     .menu-wrapper {
         display: block;
         position: fixed;
         top: 0;
-        transition: left 400ms ease;
-        left: -310px;
+        inset-inline-start: 0;
+        transition: transform 400ms ease;
+        transform: translateX(-100%);
```
StanFromIreland:
```suggestion
        transform: translateX(calc(-100% - 10px));
```
We had -310px before because of the 10px shadow.

python_docs_theme/static/pydoctheme.css
```diff
@@ -211,11 +236,16 @@ div.sphinxsidebar input[type='text'] {
 }
 
 div.body {
-    padding: 0 0 0 1.2em;
+    padding: 0;
+    padding-inline-start: 1.2em;
+}
+
+body[dir='rtl'] div.body {
+    direction: rtl;
```
StanFromIreland: No-op?

python_docs_theme/static/pydoctheme.css
```diff
@@ -233,6 +267,21 @@ div.body hr {
     height: 1px;
 }
 
+dl > dt span ~ em,
+.sig {
+    font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
```
StanFromIreland: Why is this moving?

StanFromIreland (commented):
Please update the README for this option.

sepehr-rs:
Thanks a lot for the review! I addressed your comments, please let me know if any other changes are needed.

python_docs_theme/static/pydoctheme.css
```diff
@@ -179,6 +199,11 @@ div.sphinxsidebarwrapper > ul > li > ul > li {
     margin-bottom: 0.4em;
 }
 
+div.sphinxsidebar ul ul {
```
StanFromIreland, line 202: The sidebar's rounding is incorrect:

Image

python_docs_theme/static/pydoctheme.css
```diff
@@ -233,6 +263,16 @@ div.body hr {
     height: 1px;
 }
 
+pre,
+code,
+div.highlight,
+dl > dt span ~ em,
+.sig,
+.sig-param {
+    direction: ltr;
```
StanFromIreland, line 272: This also affects signatures, no? Does that not make it inconvenient, with the propose opposite?

python_docs_theme/static/pydoctheme.css
```diff
@@ -435,6 +475,11 @@ div.footer a:hover {
     --refcount-return-borrowed-ref: var(--middle-color);
 }
 
+dl > dt span ~ em,
+.sig {
+    font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
```
StanFromIreland, line 480: Why has this moved?

README.md
```diff
@@ -22,6 +22,7 @@ To use this theme, add the following to `conf.py`:
 
 - `html_sidebars`, defaults taken from https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_sidebars
 
+- `is_rtl` - Enable right-to-left (RTL) layout for languages such as Persian, Arabic, or Hebrew. Code blocks remain left-to-right. Defaults to `false`.
```
StanFromIreland, line 25: Should be documented as a theme option, like `html_theme_options = {"is_rtl": "true"}`.

StanFromIreland (commented):
And some additional observations, footnotes and search results are still LTR:

Image
Image

python_docs_theme/layout.html
```diff
@@ -89,7 +89,7 @@ <h3>{{ _('Navigation') }}</h3>
 {%- endblock -%}
 
 {%- block body_tag %}
-{{ super() }}
```
cben: OK to drop :+1:, this file extends [classic/layout.html](https://github.com/sphinx-doc/sphinx/blob/e44a40eb2f810558ccd9da1425421270ccb81351/sphinx/themes/classic/layout.html) which extends [basic/layout.html](https://github.com/sphinx-doc/sphinx/blob/e44a40eb2f810558ccd9da1425421270ccb81351/sphinx/themes/basic/layout.html#L144) defining:

    {%- block body_tag %}<body>{% endblock %}

python_docs_theme/layout.html
```diff
@@ -89,7 +89,7 @@ <h3>{{ _('Navigation') }}</h3>
...
+<body{% if theme_is_rtl|tobool %} dir="rtl"{% endif %}>
```
cben: `super()` is OK to drop :+1:, this file extends [classic/layout.html](https://github.com/sphinx-doc/sphinx/blob/e44a40eb2f810558ccd9da1425421270ccb81351/sphinx/themes/classic/layout.html) which extends [basic/layout.html](https://github.com/sphinx-doc/sphinx/blob/e44a40eb2f810558ccd9da1425421270ccb81351/sphinx/themes/basic/layout.html#L144) defining:

    {%- block body_tag %}<body>{% endblock %}

[minor tangent]
Do users sometimes inline sphinx output into a larger page?
There can be only `body`; AFAICT browsers flatten any inner `<body>` out of the parsed DOM, and the `dir` attribute would either get lifted into the top-level `body`, or _ignored_ if the top-level body already had a dir attribute:

https://codepen.io/editor/cben/pen/01a085a5-6b2b-7e52-b131-248cf4478fa6?file=%2Findex.html%3A720%3A724&orientation=left&panel=false&show=split

So if you want to go above and beyond, could avoid confusion by opening a seperate regular div with dir="rtl", and adjusting `body[dir='rtl']` CSS rules?
**Except** basic/layout.html leaves no easy way to close `</div>` at the end, so perhaps never mind?

StanFromIreland:
Hey @sepehr-rs, just checking in. Some changes are still pending, and I'd love to land this to unblock the Persian translation. Happy to help if anything is unclear.

Also a little ping for @cben, I'd like to hear what you think. :-)

sepehr-rs:
Hi @StanFromIreland, sorry for the delay on this. I recently found a free API and ended up translating the entire Python documentation corpus into Persian ([this PR](https://github.com/python/python-docs-fa/pull/62)) and what I initially thought would be a 3-day plan turned into a 3-week project 😬.
But it’s now in the final stages of cleanup and review, so I’ll try to get back to this today or tomorrow. Thanks a lot for following up with me, I really appreciate it!

python_docs_theme/static/pydoctheme.css
sepehr-rs: Nice catch. Yes, keeping the signature itself LTR but right aligning it improves the readability :).

sepehr-rs:
Thanks a lot for your review and patience!
I’ve addressed your review comments. Please let me know if any further changes are needed.

README.md
```diff
@@ -22,6 +22,7 @@ To use this theme, add the following to `conf.py`:
 
 - `html_sidebars`, defaults taken from https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_sidebars
 
+- `is_rtl` - Enable right-to-left (RTL) layout for languages such as Persian, Arabic, or Hebrew. Code blocks remain left-to-right. Defaults to `"false"`. Set via `html_theme_options = {"is_rtl": "true"}`.
```
StanFromIreland:
```suggestion
- To lay pages out right-to-left for languages such as Persian, Arabic, or Hebrew,
  set `html_theme_options = {"is_rtl": True}`. Defaults to `false`.
```

python_docs_theme/static/pydoctheme.css
```diff
@@ -414,17 +464,48 @@ table.footnote td {
...
+    margin-inline-end: 10px;
 }
 
 div.footer a {
     text-underline-offset: auto;
 }
 
+dl > dt span ~ em,
+.sig {
+    font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
```
StanFromIreland: Can you please move this back to minimise diff? I see it moved from:

Image

python_docs_theme/static/pydoctheme.css
```diff
@@ -233,6 +269,20 @@ div.body hr {
     height: 1px;
 }
 
+pre,
+code,
+div.highlight,
+dl > dt span ~ em,
+.sig,
+.sig-param {
+    direction: ltr;
```
StanFromIreland: While these are technically fine for LTR, can we please scope it to `body[dir='rtl']` just for safety.

StanFromIreland (commented):
Overall looks good to me, last few little comments.

python_docs_theme/static/pydoctheme.css
sepehr-rs: Ugh, I found the lines and moved it back to it's place, but I think the other changes pushed it back down again. Will fix it now.

sepehr-rs:
Thanks for the additional review! Let me know if anything else needs changing.

StanFromIreland:
Please don't force push in the future, it makes reviewing more difficult.

StanFromIreland (approved):
LGTM

@hugovk, could you please take a look?

hugovk (approved):
Exciting!

@StanFromIreland Release when ready!

StanFromIreland:
Merged, thank you @sepehr-rs !

PR #326, opened 2026-08-30
Keep breadcrumbs visible on mobile

### LINKED ISSUE #317: Breadcrumb navigation visibility on mobile/narrow screens
## Summary
Breadcrumb navigation isn't visible on mobile or in narrow browser windows. This can be fixed by changing the CSS selectors slightly (instead of `div.related`, use `div.related li.right` and `div.related li.switchers` on line 568 of pydoctheme.css).

## Description
Currently, https://docs.python.org has two navigation menus in the HTML of each page, using a media query to switch between the two: 
* `<div class="related">` for desktop navigation
* `<div class="mobile-nav">` for navigation on screens (or browser windows) narrower than 1024px

The main difference between the two menus is that the `<div class="related">` menu contains breadcrumb navigation while the `<div class="mobile-nav">` menu does not. (Related: https://github.com/python/cpython/issues/123452 "Within-section navigation missing for smaller screens, global contents difficult to navigate.")

If the breadcrumb navigation was always visible, the Python docs site would be more user-friendly. 

For example, when I'm working on a project, I usually have a browser window open at the same time (side-by-side) so I can easily refer to documentation, a tutorial etc. On a laptop-size screen, I have to resize the browser window _every single time_ I want to see the breadcrumb navigation.

## Suggested change
This is the current CSS (lines 567 to 570, inside an `@media (max-width: 1023px)` media query), which removes `<div class="related">` entirely:
https://github.com/python/python-docs-theme/blob/8e9ae96a667443d50081ca067b0cdeda91d601c9/python_docs_theme/static/pydoctheme.css#L567-L570

If the selectors are changed as follows, the breadcrumb navigation remains visible on mobile while the rest of `<div class="related">` is removed:
```CSS
/* Remove sidebar, but leave breadcrumb navigation visible */
div.related li.right, div.related li.switchers, div.sphinxsidebar {
    display: none;
}
```

undeMalum:
Closes #317

---

Before the change:
image

After the change:
image


hugovk:
Thanks for the PR. Should we remove the Python logo in mobile? And possibly reduce the vertical gap between the breadcrumb lines a bit?

undeMalum:
> Thanks for the PR. Should we remove the Python logo in mobile? And possibly reduce the vertical gap between the breadcrumb lines a bit?

Now it looks like this:
image

python_docs_theme/static/pydoctheme.css
```diff
@@ -564,10 +564,17 @@ div.genindex-jumpbox a {
...
+    div.related li.right, div.related li.switchers, div.sphinxsidebar {
         display: none;
     }
+    div.related .root-icon {
+        display: none;
+    }
+    div.related ul,
+    div.related li {
+        line-height: 1.3;
+    }
```
m-aciek:
```suggestion
    div.related {
        line-height: 1.3;
    }
```

python_docs_theme/static/pydoctheme.css
```diff
@@ -564,10 +564,17 @@ div.genindex-jumpbox a {
         margin-inline-start: 1rem;
         margin-inline-end: 0;
     }
-    /* Remove sidebar and top related bar */
-    div.related, div.sphinxsidebar {
+    /* Remove sidebar and desktop-only related controls */
+    div.related li.right, div.related li.switchers, div.sphinxsidebar {
         display: none;
     }
+    div.related .root-icon {
```
m-aciek:
```suggestion
    div.related .root-icon, div.related .root-icon + li {
```
We already link to the python.org from the icon above, therefore we can drop the python.org part from the breadcrumbs completely on narrow viewports.

m-aciek (changes requested):
Thank you for the changes! Leaving two suggestions.

hugovk:
@undeMalum Please can you sign the CLA with your other email as well? 

https://github.com/python/python-docs-theme/pull/326#issuecomment-5468828902

See https://devguide.python.org/getting-started/pull-request-lifecycle/#why-do-i-need-to-sign-the-cla-again

m-aciek (approved):

m-aciek:
@hugovk @undeMalum has signed the CLA with his second email.

hugovk (approved):
Thanks!

PR #329, opened 2026-09-13
Improve contrast between unvisited and visited links in the dark mode

### LINKED ISSUE #328: Do not override the color of visited links
### Documentation

The Python documentation pages (docs.python.org) override the default color of visited links to blue / bluish purple, which makes them almost indiscernible from links not yet visited, especially in the dark theme -- see these screenshots from https://docs.python.org/3.14/library/functions.html#compile taken without any browser extensions enabled:

Image

Image
<br><br>

This is quite an oversight: when reading documentation, one wants to see which sections or pages they have already read, in order not to read them twice; and more importantly, in the case of cyclic references (which are quite common not only in the Python documentation), highlighting visited links in different color helps the reader to avoid going in circles, repeatedly visiting the same pages over and over.

For comparison, here are screenshots of the same content, but with the default color of visited links:

Image

Image
<br><br>

**Suggested solution:** Set the color of visited links to the browser default. Or, at the very least -- if you _must_ change it (which I'm pretty sure you don't) -- change it to something visibly different from the default link color.

**Temporary fix:** Use [Stylus](https://addons.mozilla.org/en-US/firefox/addon/styl-us/) or a similar extension to modify the style of `docs.python.org`, and add the following:

```css
a:visited {
    color: revert !important;
}
```

---

Browser: Firefox 153.0 (without any extensions)
OS: Windows 10 22H2

StanFromIreland:
| Before | After |
|--------|--------|
| image | image |

I think it's more distinct, while still retaining a good amount of contrast with the background (6.81).


hugovk (approved):
Thanks!

Maybe time for a release, if the RTL stuff will take a bit longer?

StanFromIreland:
> Maybe time for a release, if the RTL stuff will take a bit longer?

I'll check in on the RTL PR first.

PR #331, opened 2026-09-20
Add Vazirmatn support for Farsi

sepehr-rs:
Adds Vazirmatn for Farsi builds. [Vazirmatn](https://github.com/rastikerdar/vazirmatn) is an open-source font licensed under the SIL Open Font License. Adding this to Farsi builds improves the readability and quality of the documentation.
I didn't touch the broader RTL fonts and kept it scoped to Farsi with `:lang(fa)`. The font is bundled as a single variable WOFF2 file in `static/fonts/`, with its `OFL.txt` alongside.

With Vazirmatn:
image
image

Before:
image
image

Please let me know if any changes or additions are needed. Thanks in advance!


python_docs_theme/static/pydoctheme.css
```diff
@@ -8,6 +8,24 @@
     --bad-border: rgb(244, 76, 78);
 }
 
+@font-face {
```
StanFromIreland: Can all this be moved near the other RTL configuration, and also get a comment explaining what/why this is.

sepehr-rs: Sure, moved it next to `body[dir='rtl'] {`. Let me know if any other changes are needed!

python_docs_theme/static/pydoctheme.css
```diff
@@ -269,6 +269,28 @@ div.body hr {
     height: 1px;
 }
 
+/* Persian (fa) uses Vazirmatn, since the default system font stack renders
+   Arabic-script text poorly on many platforms. This is scoped to :lang(fa)
+   rather than [dir='rtl'] so other RTL languages (Arabic, Hebrew, Urdu...)
+   and all LTR languages are unaffected. */
```
StanFromIreland:
```suggestion
/* Persian (fa) uses Vazirmatn, since the default system font stack renders
   Arabic-script text poorly on many platforms. */
```
The rest just repeats what's below, so I suggest we drop it.

python_docs_theme/static/pydoctheme.css
```diff
@@ -269,6 +269,28 @@ div.body hr {
...
+@font-face {
+    font-family: "Vazirmatn";
+    src: url("fonts/Vazirmatn-VF.woff2") format("woff2");
+    font-weight: 100 900;
+    font-style: normal;
+    font-display: swap;
+}
+
+:lang(fa) :is(
+    body,
```
StanFromIreland: What about buttons and selects, I don't think they inherit here?

sepehr-rs:
Thanks for the review!

Sidenote: I found an issue with the sidebar while investigating this; The sidebar button's direction is reversed in RTL mode:

When closed:
image
When open:
image

I think it'd be a very small fix, but should I make a separate PR for it, or fix it here if the scope creep is acceptable?

hugovk:
Vazirmatn is unmaintained following the death of its creator:
* https://github.com/rastikerdar/vazirmatn
* https://en.wikipedia.org/wiki/Saber_Rastikerdar

Vazirharf is a maintained fork, is that suitable?
* https://github.com/nadalaba/vazirharf

In either case, the current .woff isn't too large at 109 KB, so bundling should be fine so we don't need to use a CDN. But Vazirmatn says:

> For Latin glyphs, Vazirmatn is combined with Roboto font by a build script, however there is also a version without Latin glyphs (Non-Latin).

Can we use the Non-Latin version? That'd reduce it by around half:

* Vazirmatn: I think we'd need these two:
  * 22 KB: https://github.com/nadalaba/vazirharf/blob/v34.003/misc/Non-Latin/fonts/webfonts/Vazirharf-NL%5Bwght%5D.woff2
  * 22 KB: https://github.com/rastikerdar/vazirmatn/blob/v33.003/misc/Non-Latin/fonts/webfonts/Vazirmatn-NL-Bold.woff2

* Vazirharf: just this one?
  * 50 KB https://github.com/nadalaba/vazirharf/blob/v34.003/misc/Non-Latin/fonts/webfonts/Vazirharf-NL%5Bwght%5D.woff2

sepehr-rs:
I don't oppose using a fork, but the font and its glyph coverage are stable, and the fork doesn't offer a benefit that matters for documentation (its only difference is how Lam-Alef is handled, which is aimed at typing sites). There's also no visible difference between the two, so I'd slightly lean towards staying with the original upstream release. That said, I'm happy to switch if the project prefers a maintained font, so I'll leave the decision to you.

Good point on the Non-Latin build. Vazirmatn also has a variable-weight Non-Latin version, and it's slightly smaller than Vazirharf's:
- 46.8 KB: https://github.com/rastikerdar/vazirmatn/blob/master/misc/Non-Latin/fonts/webfonts/Vazirmatn-NL%5Bwght%5D.woff2

Please let me know what you think is best going forward, and thanks for the review!

hugovk:
Let's stick with the original then, and go with that Non-Latin version. 

We can switch to a fork later if there's a concrete need to do so. Thanks!

hugovk (approved):
Thanks! I added the version number in the title, so it'll go into the commit message, so we can easily check the version.

> I think it'd be a very small fix, but should I make a separate PR for it, or fix it here if the scope creep is acceptable?

Let's do a new PR.

sepehr-rs:
Thanks a lot for the review! I'll open a separate PR for that change.

python_docs_theme/static/fonts/OFL.txt
StanFromIreland, line 1: Please add this license to `license-files` in the `pyproject.toml` file. However, from what I gather from [this discussion](https://discuss.python.org/t/expressing-project-vs-distribution-licenses-post-pep-639/90314), we don't need to list it in `licenses` as well.

sepehr-rs:
The CI is failing with `W002: Wheel contains duplicate files`. Should we ignore the warning or exclude the package copy from the wheel with `build.targets.wheel.exclude`?

StanFromIreland:
I think we can just ignore the warning?

hugovk:
Let's rather exclude, so we only have one copy in the expected place? Something this:
```toml
build.targets.wheel.exclude = [ "python_docs_theme/locale/**/*.po", "python_docs_theme/static/fonts/OFL.txt" ]
```

Ignoring W002 means we could easily end up with other duplicates in the future without realising.

sepehr-rs:
Makes sense. Excluded `OFL.txt` in `pyproject.toml`. One question though: doesn't this mean the license won't appear next to the font in the package? I thought the license should be bundled alongside the font.

hugovk:
We can see the contents of the sdist at https://github.com/python/python-docs-theme/actions/runs/35627886765

It has the licence next to the font:

```
│   ├── [2026-09-21T16:48:06Z]  static
│   │   ├── [2026-09-21T16:47:48Z]  copybutton.js
│   │   ├── [2026-09-21T16:48:06Z]  fonts
│   │   │   ├── [2026-09-21T16:47:48Z]  OFL.txt
│   │   │   └── [2026-09-21T16:47:48Z]  Vazirmatn-NL-VF.woff2
```

And it also shows the wheel, with the licence in the designated `licenses` metadata subdir:

```
└── python_docs_theme-2026.9.dist-info
    ├── METADATA
    ├── RECORD
    ├── WHEEL
    ├── entry_points.txt
    └── licenses
        ├── LICENSE
        └── python_docs_theme
            └── static
                └── fonts
                    └── OFL.txt
```

But it doesn't need to be directly next to the font. The licence says:

> 2) Original or Modified Versions of the Font Software may be bundled,
redistributed and/or sold with any software, provided that each copy
contains the above copyright notice and this license. These can be
included either as stand-alone text files, human-readable headers or
in the appropriate machine-readable metadata fields within text or
binary files as long as those fields can be easily viewed by the user.

* "... provided that each copy contains the above copyright" ->  "Copyright 2015 The Vazirmatn Project Authors (https://github.com/rastikerdar/vazirmatn)"

* "and this license" -> check

* "These can be included ... in the appropriate machine-readable metadata fields within text or binary files" -> using the wheel metadata for licences in the [Core Metadata Spec](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata) defined by [PEP 639](https://peps.python.org/pep-0639/).

sepehr-rs:
Thanks a lot for taking the time to explain this to me! I understand it now =).

python_docs_theme/static/pydoctheme.css
```diff
@@ -269,6 +269,28 @@ div.body hr {
     height: 1px;
 }
 
+/* Persian (fa) uses Vazirmatn, since the default system font stack renders
+   Arabic-script text poorly on many platforms. */
+@font-face {
+    font-family: "Vazirmatn";
```
StanFromIreland:
```suggestion
    font-family: "Vazirmatn NL";
```

python_docs_theme/static/pydoctheme.css
```diff
@@ -269,6 +269,28 @@ div.body hr {
...
+}
+
+:lang(fa) :is(
+    body,
+    div.body :is(h1, h2, h3, h4, h5, h6),
+    div.sphinxsidebar :is(h3, h4),
+    input,
+    select,
+    textarea,
+    button:not(.copybutton)
```
StanFromIreland: This all seems a little duplicative, maybe just:
```
    body,
    input,
    select,
    textarea,
    button:not(.copybutton),
```
should work?

python_docs_theme/static/pydoctheme.css
```diff
@@ -269,6 +269,28 @@ div.body hr {
...
+) {
+    font-family: Vazirmatn, -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif;
```
StanFromIreland: Since we now duplicate the stack in three places, can you please move it to `root` as `--sans-serif-font-stack`.

python_docs_theme/static/pydoctheme.css
sepehr-rs: Actually no, I had to make it even more specific; otherwise, the font wouldn’t apply to some elements. Several theme rules set `font-family `directly on headers and sidebar/related inputs, which overrides inheritance from `body`. So I had to target those elements directly to override those rules.

